PR #10, opened 2017-01-10
Initial embedded discuss comments

kallisti5:
Initial work on embedded discuss comments on blog posts


layouts/partials/discourse.html
```diff
@@ -0,0 +1,13 @@
+<!-- layouts/partials/discourse.html -->
```
waddlesplash: Unnecessary comment?

avongluck-r1soft: Not really.. while we're debugging the comments it is nice to know the partial actually loaded.

PR #16, opened 2017-01-17
Initial pages for GSoC 2017

pulkomandy:
(I would push directly, but it seems I'm not allowed?)


waddlesplash:
Looks like GitHub changed their permission system yet again. Fixed.

PR #17, opened 2017-01-20
css: Tweak page min-width size to 1100px

kallisti5:
* The minimum user screensize in 2014 was 1024x768 at
  3.7% of web broswsers (mobile + non).
* This change means the minimum "desktop" resolution to
  display our site is 1024x768
* The site already dynamically scales to a mobile view
  on smaller screens.
* This should give us more room for content making our
  site look less constricted and antiquated to most users.

### DIFF
--- a/static/css/shijin4.css
+++ b/static/css/shijin4.css
@@ -29,7 +29,7 @@ a {
 }
 
 #page {
-	max-width: 965px;
+	max-width: 1100px;
 	margin-left: auto;
 	margin-right: auto;
 }


pulkomandy:
We should take into account that people don't necessarily run Web+ fullscreen, and on a 1280x768  display (typical 16:9), if you leave space at the right for the Deskbar, your window is about 1024px wide (coincidence?).

Using global stats which don't take the specificities of our userbase into account may not be the wisest choice.

Are flexible layouts still not possible in modern CSS?

richienyhus:
@pulkomandy I just suggested [Unsemantic](http://unsemantic.com) on the Haiku-Web ml.

>Unsemantic is a fluid grid system that is the successor to the 960 Grid System. It works in a similar way, but instead of being a set number of columns, it's entirely based on percentages.

avongluck-r1soft:
The fixed width page is a bit old-school.  We could do a lot of awesome things like unsemantic.  The new site does adjust to different layouts unlike the old one.  The fixed width comes from our original theme. This change just makes things a "little less bad" until we have time to refactor the layout long-term.

waddlesplash:
> The fixed width page is a bit old-school.

Well, as you said yourself, it's not exactly fixed-with, it's "max-width". Personally, I find the column max size quite nice, and better for readability (lines are not too long before wrapping) -- so let's have this bikeshed later.

PR #25, opened 2017-01-26
Layout slideshow for Desktop and Mobile

jscipione:
Using a "responsive" design. Replaces table with list that works like a table
on Desktop and a list on mobile.

### DIFF
--- a/content/slideshows/haiku-2.html
+++ b/content/slideshows/haiku-2.html
@@ -30,7 +30,7 @@ hr { border: 1px solid #D6DDB9; height: 1px; background: #D6DDB9; margin: 6.6px 
 	<img src="../../files/apps-1_preview.png" width="576" alt="StyledEdit, DeskCalc, DiskUsage, Terminal" /></a></li>
 	<!-- Showing StyledEdit, DeskCalc, DiskUsage, Terminal, Expander -->
 	<li><hr /><p class="caption">"A simple text editor, calculator, disk space monitor, un-archiver and Haiku's bash shell."</p>
-	<p class="further">Further reading: <a href="/docs/userguide/en/applications/stylededit.html">StyledEdit</a>, <a href="/docs/userguide/en/applications/deskcalc.html">DeskCalc</a>, <a href="/docs/userguide/en/applications/driveusage.html">DriveUsage</a>, <a href="/docs/userguide/en/applications/expander.html">Expander</a>, <a href="/docs/userguide/en/applications/terminal.html">Terminal</a></p><hr /></li>
+	<p class="further">Further reading: <a href="/docs/userguide/en/applications/stylededit.html">StyledEdit</a>, <a href="/docs/userguide/en/applications/deskcalc.html">DeskCalc</a>, <a href="/docs/userguide/en/applications/diskusage.html">DiskUsage</a>, <a href="/docs/userguide/en/applications/expander.html">Expander</a>, <a href="/docs/userguide/en/applications/terminal.html">Terminal</a></p><hr /></li>
 </ul>
 <br />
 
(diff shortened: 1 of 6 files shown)


content/slideshows/haiku-2.html
```diff
@@ -16,74 +16,60 @@
...
+	<p class="further">Further reading: <a href="/docs/userguide/en/applications.html">Applications</a>, <a href="/docs/userguide/en/desktop-applets.html">Desktop Applets</a>, <a href="/docs/userguide/en/preferences.html">Preferences</a></p><hr /></li>
+</ul>
 
-<tr><td height="15"></td><td></td></tr>
+<ul class="table">
+	<li><a target="_blank" href="../../files/apps-1.png">
+	<img src="../../files/apps-1_preview.png" width="576" alt="StyledEdit, DeskCalc, DiskUsage, Terminal" /></a></li>
+	<!-- Showing StyledEdit, DeskCalc, DiskUsage, Terminal, Expander -->
+	<li><hr /><p class="caption">"A simple text editor, calculator, disk space monitor, un-archiver and Haiku's bash shell."</p>
+	<p class="further">Further reading: <a href="/docs/userguide/en/applications/stylededit.html">StyledEdit</a>, <a href="/docs/userguide/en/applications/deskcalc.html">DeskCalc</a>, <a href="/docs/userguide/en/applications/driveusage.html">DriveUsage</a>, <a href="/docs/userguide/en/applications/expander.html">Expander</a>, <a href="/docs/userguide/en/applications/terminal.html">Terminal</a></p><hr /></li>
```
humdingerb: can you change
 ```
<a href="/docs/userguide/en/applications/driveusage.html">DriveUsage</a> to
 <a href="/docs/userguide/en/applications/diskusage.html">DiskUsage</a>
```
while you're at it?

jscipione: Sure, I have to update this a bit more anyway

jscipione:
I have updated this pull request so that it works appropriately including a couple of edge cases without causing issues elsewhere on the site. I now recommend this pull request to be applied by the powers that be. The intended result is that the slideshow works on many different page sizes including mobile ones. Tested on Google Chrome using Inspector tool.

I also implemented humdinger's suggested correction, thanks for noticing the issue.

humdingerb:
> I also implemented humdinger's suggested correction, thanks for noticing the issue.

That was Richie's doing, he reported it in https://dev.haiku-os.org/ticket/13252
I'll let waddlesplash or kallisti5 merge your change, as I have no clue about those things...

jscipione:
Updated commit to also thank Richie

waddlesplash:
Manually merged in f22f53e with a few tweaks of my own.

jscipione:
Slideshow page looks good, but, unfortunately your tweaks have undermined my indented fix which allows layout issues to persist elsewhere. For example on the home page the Haiku Logo and Download it now! GET HAIKU images are expanding the page width on mobile. This is pretty minor breakage since the widths of those images are only slightly wider than the page width, but, if you look at another page, fox example on the Virtualizing Haiku page, the images are much wider and thus have a much more detrimental affect on the layout. I will attach some screenshots showing the problems.

homepage mobile layout problems
more dramatic article page layout problems

waddlesplash:
I figured that if the screenshots page was broken, with that CSS, then I figured others could be too, so I didn't leave it enabled for all pages till we can audit those too.

jscipione:
Exporing the img { max-width: 100% } code from the slideshow page to all pages should fix the issue unless the images have a specified a height. If any images do have a specified height I would consider that to be a bug as the height should expand with the width. One way to get around that would be to specify img { height: auto !important; } to override any images that have specified a height. Wrap all that in a @media (max-width:767px) block so that it only applies on mobile.

PR #27, opened 2017-02-05
Added Graph

deep-1:


humdingerb:
I already moved the graph. I did the change in the text in spirit, but didn't use exactly your words. I hope you don't mind.

It looks like using HTML is a safer bet than markdown. There were some formatting issues...

deep-1:
The blog post looks good to me now . Thanks !

PR #28, opened 2017-02-08
Reset fundraising themometer for 2017

wezm:
Addresses [my comment](https://dev.haiku-os.org/ticket/11154#comment:20) on bug [#11154](https://dev.haiku-os.org/ticket/11154):

> Can we get the Fundraising heading on the website home page updated and the the amount reset?
>
> "Fundraising 2016", is one of the most prominent headings above the fold when you land on the home page and the reference to 2016 makes it feel out of date.

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -47,7 +47,7 @@
 			</div>
 		</div>
 		<div class="col-md-4" id="preface-last">
-			<h2 class="fundraising">Fundraising 2016</h2>
+			<h2 class="fundraising">Fundraising 2017</h2>
 			<div class="content">
 				<div style="margin-top: 30px;">
 					<div style="background-image:url(/files/images/piggy-bank.png);background-repeat:no-repeat; background-position: left; font-size: 0.9em; padding-bottom: 8px;">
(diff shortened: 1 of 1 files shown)


leavengood:
If you want the real numbers here is what they currently are:

Total for EUR: 205.00
Total for USD: 512.00
Grand Total (at EUR to USD rate of 1.066500): 730.63
Percentage pixels (from 128): 2.67

leavengood:
I have a program which gets this information from PayPal, we just need to figure out how we want to automate it to update the website daily or weekly or whatever. I'm tired of doing it manually.

leavengood:
I also can track the addition and removal of subscriptions if we want to show those.

avongluck-r1soft:
@leavengood   can we put it on one of the servers?  Since the site is in git now, we could automate a daily git commit pretty easily.

kallisti5:
@leavengood can we put the code on one of the servers?  We/I could automate updating the page and checking it in pretty easily.

wezm:
I have updated with the figures provided by @leavengood 

image

kallisti5:
@wezm   looks good!  I'd merge but I don't have permissions :-(

leavengood:
Apparently I can merge, so I did.

@kallisti5 yeah putting my code on a server would be nice. I just need to adjust it to fill out the HTML rather than outputting the information as it does now.

PR #30, opened 2017-02-20
Fix missing word on developing page

kainjow:


waddlesplash:
Thanks!

waddlesplash:
It looks like your changes altered the filemode (100644 => 100755). I fixed it in 8a55b869dd1e4b7d7d91105b58ffdcbeef34a64e, so just a heads-up to watch out for that next time.

kainjow:
Ah, my bad. Do all files need the same permissions? Could chmod them all in the deployment script. That's what I do for my Jenkins sites.

waddlesplash:
> Do all files need the same permissions? Could chmod them all in the deployment script. That's what I do for my Jenkins sites.

Uh, no, you should just commit them with the right filemodes. Why are files you change or add getting their `x` bit set anyway?

kainjow:
Probably the text editor I was using. I certainly didn't manually change it.

It's weird that GH's diff for the patch doesn't show the file mode change, but your commit does.

kainjow:
Aactually, I think I used GitHub's web editor to make the change, which uses the -patch branch name, so maybe it's a GH bug?

kainjow:
Ok, it was the c4ccaf7 that changed it. Not sure how that happened, I'll try to watch for it next time.

PR #31, opened 2017-02-20
Escape code that contains brackets

kainjow:
Fixes Trac [#13319](https://dev.haiku-os.org/ticket/13319)

How it currently looks:

old

How it'll look with this fix:

new


waddlesplash:
I suppose this works for now.

PR #34, opened 2017-02-21
Add missing conference (Upcoming Events) index page

kainjow:
Fixes Trac [#13237](https://dev.haiku-os.org/ticket/13237)

The page uses a shortcode which takes events from a data file and dynamically outputs current and past events, based on the event’s end date.

I had to use a shortcode because Hugo does not allow template code to be used in content files. Also I had to use Hugo's Scratch object because Go templates do not allow reassigning a variable in another scope (Jekyll can do both of these things :).

So essentially the data file only needs to be modified, and when the site is rebuilt it will dynamically recreate the upcoming and past events. If there are no current events it will show "No events available." This could also be done live through JavaScript but that seems a little heavy.


waddlesplash:
> I had to use a shortcode because Hugo does not allow template code to be used in content files. Also I had to use Hugo's Scratch object because Go templates do not allow reassigning a variable in another scope (Jekyll can do both of these things :).

I think this can be done without a .yml file by just treating the "conference" subdir as a taxonomy (vaguely like what I did for the "blog" index, which Hugo treats a similar way.) That should get rid of the shortcode altogether.

If you want to try your hand at doing it that way, please do; otherwise I'll look into it sooner or later.

waddlesplash:
> (Jekyll can do both of these things :).

While I could rant quite a lot about Go's arcane and backwards template syntax, the fact remains that it's almost universally better than the competition in so many ways. I took a brief glance at Jekyll, but from what I recall, it was written in Ruby (and thus slower as well as a giant pain in the neck to install. I mean, seriously, a lot of the reason to go with Hugo was that it was a single binary that you copied to your local bin dir, and that it was written in Go which means it can build the entire Haiku site in ~7s.)

If I had all the time in the world, I'm sure I would have invented yet another static site engine "the Haiku way", but as I didn't, Hugo it was. :)

kainjow:
Ok I'll try out the taxonomies. Definitely seems a better idea as it'd consolidate all event date into a single file, and easier for new contributions as they can just create one of those pages and submit as a PR.

Eh, if you've never used Jekyll/Liquid to build a large site, hard to say Go is better. I've done a few sites with it and yes it is slow (I'm experimenting with a port of it to C++ to solve this) but from the little I've done with Go/Hugo I already feel constrained compared to Liquid. I also agree distribution of a binary is far better than managing gems (try explaining that to a non-techie). Still, good to learn another way of doing things :)

kainjow:
Did a conversion to taxonomy. Can't seem to get the list to sort by the event end date as Hugo's sort function doesn't allow sub keys, so right now it's just sorting by Date, which is generally good enough.

If you think this direction is good I can continue adding the front matter to the rest of the files.

layouts/indexes/conference.html
```diff
@@ -0,0 +1,92 @@
...
+					{{ end }}
+
+				{{ end }}
+
+				</table>
+				
+				</p>
+			</div>
+		
+	</div>
```
waddlesplash: Lots of extraneous blank lines -- you can tidy those up a bit.

waddlesplash:
Yep, looks very good. Adding all the rest of the front matter seems like a good idea. Thanks for doing this!

PR #35, opened 2017-02-22
_redirects: Try cross-domain redirects.

waddlesplash:
Don't mind me...


waddlesplash:
It doesn't work. Oh well.

PR #37, opened 2017-02-23
Add sidebars to About, Community, Development and Documents pages

kainjow:
I first tried creating the menus dynamically by grabbing all top-level files in each sub directory (going off the first component of the URL), but this showed unwanted pages. Also several menus include pages that don't share the same base URL, like Documents. So I just reverted to generating them manually, which gives far more control and is simpler.


pulkomandy:
I would not add the "documents" one in that shape. We should re-think the organization of that section completely, with "/guides" being the main entry point, and the sidebar probably matching its contents.

Otherwise, +1.

xdev31:
+1

kainjow:
Looks like this has fallen off the radar. Any further feedback?

PR #46, opened 2017-05-05
Fix blog index not working due to case sensitive compare

kainjow:


pulkomandy:
Thanks for watching :)

PR #53, opened 2017-05-08
GSoC 2017: Calendar Application intro blog post

AkshayAgarwal007:


humdingerb:
Maybe  @waddlesplash can check that page before merging it... But past experience showed that our "discourse" doesn't like some markdown formatting. It may be a better idea to go with HTML instead.

PR #127, opened 2017-10-09
kallisti5: where is r1? post

kallisti5:


nielx:
+1 from me (fwiw)

korli:
I thought first the hostname was with underscores \_maui\_ , then realized it's markdown :)

Otherwise a typo "virtulization". 

On a side node, it would be nice to have the buildbot config synched from a git repo (see https://dev.haiku-os.org/ticket/13311)

PR #128, opened 2017-10-14
r1b1-notes: clarification to variants

kallisti5:


waddlesplash:
Clarified this slightly in my offline copy.

PR #130, opened 2017-10-19
Add Events (Capitole du Libre; Alchimie 12)

mmuman:


mmuman:
I wasn't sure about keeping HTML format, so I had a try at markdown, doesn't look too bad I think.

kallisti5:
lgtm

PR #135, opened 2017-10-20
Add a Packages activity tab

mmuman:
Not sure it will work on the real site, due to CORS policy...


waddlesplash:
It won't, but we use Netlify proxying for the ML tab already. See how the `/exapi/` stuff works in the code -- and the server-side config is: https://github.com/haiku/website/blob/master/static/_redirects

mmuman:
Ah yes I saw this but didn't look much. Should work now.

PR #139, opened 2017-11-09
about: Add maintainers blurb

kallisti5:
We spoke about this on the mailing list.
https://www.freelists.org/post/haiku-commits/haiku-hrev51452-srcappsaboutsystem,2

### DIFF
--- a/content/about/index.md
+++ b/content/about/index.md
@@ -31,3 +31,20 @@ Haiku is developed by a growing community of volunteer developers from all aroun
 
   * [Discover the Haiku Community](/community)
   * Check out [Haiku Contributors](https://github.com/haiku/haiku/graphs/contributors) at GitHub
+
+### Who are the maintainers in the About window?
+
+Active developers are considered "Maintainers" of the Haiku operating system.
+
+  * Anyone who has contributed directly to the haiku or haikuporter codebase within the last 24 months.
+  * Anyone who has contributed directly to haikuports 12 or more times in the last 12 months.
+
+Changes to the maintainers list are done by hand and groomed to the best of our ability.
+If you feel left out (or unfairly included), let us know via the mailing lists!
+
+Example usage of git to check authors since December 2016:
+```
+git log --after=Dec2016 | grep Author: | sort | uniq -c | sort -n
+```
+
+Maintainers are moved to "Past Maintainers" once the conditions above are no longer true.


content/about/index.md
```diff
@@ -31,3 +31,20 @@ Haiku is developed by a growing community of volunteer developers from all aroun
 
   * [Discover the Haiku Community](/community)
   * Check out [Haiku Contributors](https://github.com/haiku/haiku/graphs/contributors) at GitHub
+
+### Who are the maintainers in the About window?
+
+Active developers are considered "Maintainers" of the Haiku operating system.
+
+  * Anyone who has contributed directly to the haiku or haikuporter codebase within the last 24 months.
+  * Anyone who has contributed directly to haikuports 12 or more times in the last 12 months.
```
waddlesplash: 1 contribution a month is not exactly "maintainer" ... how about we make it 32 contributions over the past 12 months? Then it's a nice power of 2.

korli: Also with commit rights.

kallisti5: I'm strictly going off of what was discussed on the mailing list. What if someone makes 1 commit of 10,000 lines?

kallisti5: Once we migrate to gerrit, the language becomes easier.  "Anyone with commit access, or who has submitted and had 16 patches accepted"?

jt15s: Any word on this? It's been a while (3+ years to be exact) since this was opened...

pulkomandy:
Similar information has since been added to the "governance" page: https://www.haiku-os.org/community/organization/governance/

The version in this merge request summarizes an old discussion about when to remove people from the active maintainers list, but as far as I know this was never actually deployed. We have several people not having contributed for 3+ years and still listed as maintainers. And number of commits is not a good measure, some people don't write any code but are still members of the team for other reasons (QA testing, for example).

So the current page only says what's agreed on: how to get into the list. There is currently no clear way to say when people get removed from there.

leavengood:
Since this has gone quite stale I am closing it. If we want to revisit it someone can reopen, but as jt15s says it has been more than 3 years so clearly there is not a lot of interest here.

I feel like we have an intuition when someone should be removed from the Maintainers list. And if one of those people comes back they can always moves themselves back. I did this in 2019 I think.

PR #147, opened 2017-12-04
[GCI2017] Rewrite the Parallels Desktop guide

Damillora:
The rewritten guide is written with the style of the VirtualBox guide, because I think the VirtualBox guide's style is easier to understand than the current Parallels Desktop guide.

The workaround is still needed and I've rewritten the guide with that in mind.


humdingerb:
Sensible to have adopted the style of existing guides. Looks good! Can't check, of course, but does look good... :)

PR #152, opened 2017-12-08
Add Xen virtualizing guide

Damillora:
I actually reused the VM creation part from the KVM guide, with some updates, because the Xen guide also uses virt-manager like the KVM guide.

Made as part of Google Code In 2017


humdingerb:
Thank you.

PR #162, opened 2018-01-09
PermitRootLogin yes

tangorri:
Was indicating wrong value for `PermitRootLogin`


waddlesplash:
Thanks for the patch!

PR #164, opened 2018-01-25
funding: add my liberapay profile.

fbrosson:


fbrosson:
I hope this is OK :)
Thanks!

(I did not dare to update the timestamp in line 4. Should I do so?)

PR #171, opened 2018-03-08
newblog script: sync after transition from "index.md" to "_index.md"

fbrosson:
Ref: 8264ce92a4eb9e7bfa7ab402315b34853959df97

### DIFF
--- a/scripts/newblog.sh
+++ b/scripts/newblog.sh
@@ -20,5 +20,5 @@ fi
 
 DIRNAME=`echo $1 | tr '[:upper:]' '[:lower:]'`
 mkdir -p content/blog/$DIRNAME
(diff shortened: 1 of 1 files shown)


fbrosson:
Should I also replace `index` by `_index` in lines 167, [169](https://github.com/haiku/website/blob/master/scripts/drupal2hugo.js#L169) and 171 of `scripts/drupal2hugo.js` ?

fbrosson:
I guess the `scripts/drupal2hugo.js` script was only used for the migration.
(That's why I did not change it.)

waddlesplash:
Indeed it was; I'm keeping it for historical purposes (and maybe someone else may use it someday, although it'll need modification to work with a specific Drupal install and newer Hugo.)

Thanks for this catch, I totally forgot about it!

PR #173, opened 2018-03-21
Added Updating Haiku to Documents Page

digib0y:
Fixes https://dev.haiku-os.org/ticket/13674 which is assigned to @humdingerb and was reported by @Perelandra0x309


humdingerb:
Maybe title the link "Updating and downgrading Haiku" instead.

waddlesplash:
Eh, fine as is.

PR #176, opened 2018-03-25
get-haiku: Finally clean up old r1a4, drop dead mirror links

kallisti5:


kallisti5:
get-haiku

kallisti5:
The release dates might need to be checked. I used the timestamps on the mirrors.

PR #187, opened 2018-05-08
pre-reqs: add mtools as a required package

jessicah:
Once the UEFI support is added to the anyboot image, mtools will be a required dependency. Update the documentation, moving mtools from ARM requirements to main requirements on supported platforms.


korli:
Jessica, on Haiku, I see mtools is only for x86_64 available:
https://github.com/haikuports/haikuports/blob/master/sys-fs/mtools/mtools-4.0.18.recipe
Could this be a problem when the prereq is added in configure script?

PR #218, opened 2018-07-09
URL with https

zeilen:
Old URL uses http and different host. New URL uses https and same host.

### DIFF
--- a/themes/shijin4/layouts/index.html
+++ b/themes/shijin4/layouts/index.html
@@ -70,7 +70,7 @@
 					<ul class="fundraising-details">
 						<li>Updated: February 10th</li>
 						<li>€1 = $1.2273</li>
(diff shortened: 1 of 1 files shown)


kallisti5:
I'm consulting the Haiku, Inc. board on this one.    The donation page @ haiku-inc.org is the "official" one.  I wasn't even aware of the copy on our site.

I'm mostly thinking in terms of potential for tampering with the one on haiku-os.org since it is in git and we do a lot of pull requests. (your change is 100% ok... just thinking about other potential PR's :-) )

kallisti5:
One potential is to get haiku-inc.org into another git repo, then convert it to Hugo as well.  Then only board members could update the donation links.

zeilen:
please mention, the top-links piggybank already links to there

waddlesplash:
I've "merged" this change manually, but deleted the page it points to and replaced it with a redirect to haiku-inc.org, which now uses HTTPS.

PR #229, opened 2018-08-02
Donation meter updates

leavengood:
Work to make the donation meter stay more up-to-date.

I've moved the donation meter HTML into a partial, updated it to move out most of the inline CSS, and then added JavaScript to update it based on a JSON file.

The intent will be to have my Go donation meter program update the JSON once a day.

Worse case if the JavaScript fails, we still have the most recent HTML update from August 1, and I could see updating it once a month or so.

I think I've kept to pretty basic JavaScript, and this works on Firefox, Chrome and Safari on macOS High Sierra. I was not able to test WebPositive in Haiku.

The style of the meter change a bit, here is what it looks like now:

screen shot 2018-08-02 at 12 17 30 pm

### DIFF
--- a/themes/shijin4/layouts/partials/fundraising.html
+++ b/themes/shijin4/layouts/partials/fundraising.html
@@ -1,0 +1,28 @@
+<h2 class="fundraising">Fundraising 2018</h2>
+<div class="content">
+  <div style="margin-top: 30px;">
+    <div class="fundraising-meter" style="width: 260px; min-height: 126px; max-width: 100%; background-image:url(/files/images/piggy-bank.png);background-repeat:no-repeat; background-position: left; font-size: 0.9em; padding-bottom: 8px;">
+      <div style="position:relative;
+            border-right: 1px solid #060606; border-bottom: 1px solid #060606;
+            opacity: 0.9; left:96px; width: 20px; height:128px;">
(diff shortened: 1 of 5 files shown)


leavengood:
Also FWIW this is what the JSON looks like:

````json
{
  "updated_at": "2018-08-02T11:35:43.511Z",
  "usd_donations": 4213.77,
  "eur_donations": 1146,
  "eur_to_usd_rate": 1.160950,
  "total_donations": 5544.22
}
````

File: themes/shijin4/layouts/partials/fundraising.html [hunk 1]
```diff
@@ -0,0 +1,21 @@
+<h2 class="fundraising">Fundraising 2018</h2>
```
waddlesplash: This file should be in the top level "layouts" directory, as it's not theme-specific.

themes/shijin4/static/css/shijin4.css
```diff
@@ -194,6 +194,48 @@ nav.navbar button.btn:hover {
 	margin-top: 0px;
 }
 
+.fundraising-meter {
```
waddlesplash: Since these are only used on the frontpage and are not theme specific, they should go in an inline <style> tag in the partial file instead of here.

leavengood:
OK, fixed your comments.

PR #238, opened 2018-10-12
minor additions for helping newbies

theclue:
I would like to propose a little addition on the guide, after spending a couple of hours yesterday trying to figure out how to configure GRUB. I think that newbies just like me would benefit on having more explicit steps to follow and a link to the official UEFI installation guide.

Changes:

* Example of commands for having UUID data under Linux
* Link to the official UEFI guide and fix the chainloader path accordingly in menuentry
* Installation of Grub Configurer, as it could help newbies
* Removed R1B1 from entries so the guide will be stay forever after (no need to specify version IMHO, hope this is ok)


theclue:
Tnx allot. Watch out I've just noticed that the brakets in the in link to the UEFI guide are backwarded, so the Markdown link doesn't work. Sorry!

humdingerb:
For the life of me, I cannot find what's wrong with:
`[in the official guide](/guides/uefi_booting)`
It's `[title](url)`...
/me scratches head

humdingerb:
Correct! Thanks, fixed now.

PR #239, opened 2018-10-16
Fixed broken link

victordomingos:


waddlesplash:
We really just need to fix the redirect. Paging @kallisti5...

korli:
I think the redirect is correct, https://api.haiku-os.org/ => https://www.haiku-os.org/docs/api/ because 
HIG isn't really a part of API.

PR #243, opened 2018-10-25
Add tip to use --depth when cloning

humdingerb:


content/guides/building/get-source-git.md
```diff
@@ -45,6 +45,7 @@ git clone git://git.haiku-os.org/haiku
 -->
 </li>
 </ul>
+<p>If you don't care about the commit history or want to keep the download small, try using the parameter <tt>--depth</tt> when cloning. <tt>--depth 10</tt> limits the history to the last 10 commits, for example.</p>
```
waddlesplash: how about "and want" instead of "or want" here?

PR #245, opened 2018-11-14
events: Add Capitole du Libre 2018

mmuman:
It's a copy-paste from last year. Maybe @pulkomandy would have some changes to make.


waddlesplash:
Looks OK to me.

PR #247, opened 2018-11-20
Pre reqs: homebrew/dupes/less->less

MRNIU:
At https://www.haiku-os.org/guides/building/pre-reqs, OS X

    brew install autoconf cdrtools gawk wget nasm homebrew/dupes/less mpfr gmp libmpc bison mtools

`homebrew/dupes/less` seems to be no longer usable, here is brew's log

    Error: homebrew/dupes was deprecated. This tap is now empty as all its formulae were migrated.

I use `brew install less` to install `less`.


waddlesplash:
Thanks!

PR #249, opened 2018-11-22
Change "Get Haiku" badge on main page

humdingerb:
A while back I did a modified "Get Haiku" badge for the beta
release. I changed it a bit and propose it here once more, esp.
after PulkoMandy pointed out that having bugs crawling around
isn't that inviting... :)

Here's a preview: https://0x0.st/s0rM.png


waddlesplash:
🎊

PR #252, opened 2018-12-24
Usb update

Danc2050:
Modified USB install:
- Added a small amount of text
- Inserted a picture for the Installer step.


content/guides/installing/making_haiku_usb_stick.html
```diff
@@ -61,7 +61,9 @@ <h1><a id="#haiku">Haiku: Using Installer and DriveSetup</a></h1>
 
 <p>The last step is to go to <span class="menu">Partition | Format | Be File System...</span>, choose a name and format the new partition.</p>
 
-<p>Close DriveSetup and launch Installer.</p>
+<p>Close DriveSetup and launch Installer (found under <span class="menu">Applications</span> in the Deskbar).</p>
```
waddlesplash: All applications are found in the Deskbar. What is the use of this aside?

content/guides/installing/making_haiku_usb_stick.html
```diff
@@ -61,7 +61,9 @@ <h1><a id="#haiku">Haiku: Using Installer and DriveSetup</a></h1>
...
+
+<p><img src="/files/guides/installing/usb_installer_onto.png" alt="Install Haiku Onto the new partition" /></p>
```
waddlesplash: I'm also not sure what value this image is...

content/guides/installing/making_haiku_usb_stick.html
Danc2050: The same aside is mentioned previously about the DriveSetup earlier, so I was thinking it could be misleading to not also mention where the Installer was located...could be wrong though.

content/guides/installing/making_haiku_usb_stick.html
Danc2050: All other steps are documented with pictures besides this final (and important) step. If a user is trying to install quickly or switching back and forth between screens often, I hypothesize they potentially could miss this final step and believe the instructions did not work.

content/guides/installing/making_haiku_usb_stick.html
Danc2050: Additionally, the same aside is mentioned earlier in the text... which makes this line a repeat, so there is probably not much point to this particular commit.

humdingerb: I put the "(found under Applications in the Deskbar)" when mentioning DriveSetup, because in my experience quite a few people imagine it in "Preferences". I kinda understand why, because in a way you change the configuration of your system...

Anyway, I wouldn't mind having the same for the Installer here, for that same reason. OTOH, it may just suffice to point it out once for DriveSetup and trust that the user will return to the Applications menu to find all susequently mentioned applications.
Hmm... I tend to prefer mentioning it only once...

content/guides/installing/making_haiku_usb_stick.html
humdingerb: I agree, @Danc2050.

PR #255, opened 2019-01-11
Fix incorrectly terminated list & terminal section

rrobgill:
The "some notes" list was terminated early, and blank line prior to terminal markup preventing it from rendering correctly.

image


waddlesplash:
Thanks!

PR #256, opened 2019-01-15
added python to build system requirements

aplgithub:
It has been pointed out that my [change](https://review.haiku-os.org/c/haiku/+/858) is introducing python as a build dependency.  This PR updates the dependencies page as suggested.  This just covers those platforms that don't have python by default.  I _think_ those packages' names are correct for python on those platforms, but am not sure as I don't have test hosts.

### DIFF
--- a/content/guides/building/pre-reqs.md
+++ b/content/guides/building/pre-reqs.md
@@ -60,7 +60,7 @@ sudo apt-get install u-boot-tools util-linux device-tree-compiler bc
 **Basic requirements:**
 
 ```sh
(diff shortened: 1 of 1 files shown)


korli:
For ubuntu, it's "python3" https://wiki.ubuntuusers.de/Python/#Python-3-x

aplgithub:
Thanks for the tip on Ubuntu, but I think Python comes out of the box on that distribution so there is no need to add it to the packages list.

korli:
Well it is always possible that a default package was uninstalled. I think you're overthinking, the command line should just list the required packages.

PR #257, opened 2019-01-15
Font used on tabs to work on hosts without 'Verdana' font installed

aplgithub:
The font "Verdana" is not installed on Ubuntu (for example) by default so the tabs are appearing as the default serif fonts on Ubuntu in Firefox which doesn't look too good.  I'm not sure why the tabs on the home page are in Verdana; so this PR changes those to the same fonts as the rest of the page.


mmuman:
Why remove Verdana from the list? You can just add the others, the first installed will be used.

aplgithub:
Hello @mmuman ; I though it must be a mistake.

PR #260, opened 2019-01-24
asciinema: Add support for asciinema casts (they're awesome)

kallisti5:


themes/shijin4/layouts/partials/footer.html
```diff
@@ -2,5 +2,6 @@
 	<link href="https://fonts.googleapis.com/css?family=Open+Sans:400,700" rel="stylesheet">
 
 	<script src="/js/navbar.js"></script>
+	<script src="/js/asciinema-player.js"></script>
```
waddlesplash: This file is probably pretty large; please only include it when it's actually needed. Same for the CSS.

jt15s:
Is this still required? Looking at the website here: https://asciinema.org/ it seems this is a tool for recording terminal sessions, which _could_ come in handy for documentation, but would be good to see everyone else's opinion on this.

pulkomandy:
I don't see why we need it. Looks like kallisti5 thought it looked cool, but we are not really a terminal-oriented system so we don't really have a need for this.

For the occasional list of commands in documentation, a static view seems to be better. This tool is mostly useful for demonstrating ncurses apps or other similar complex things running in Terminal, but we don't have a lot of that here.

jt15s:
Ah ok. In that case, should someone close this PR?

PR #269, opened 2019-02-28
Fix typo in the BeBook

CodeforEvolution:
Change BTimeSource to BSoundPlayer within the BSoundPlayer documentation.


pulkomandy:
Unfortunately, the Be Book was offered to us by Access Co, Ltd. only under a BY-NC-ND license. The ND part means we are not allowed to make changes to it, even fixing typos.
I suggest writing new documentation in the Haiku book at http://api.haiku-os.org instead.

PR #271, opened 2019-03-05
Update KVM instructions for Ubuntu 18.10 and later

glesica:
I noticed that the KVM installation instructions no longer apply to Ubuntu 18.10 and later, so I figured I would update them.


waddlesplash:
Thanks!

PR #280, opened 2019-04-16
news: NVMe driver in tree

kallisti5:


pulkomandy:
Well... did we give up on the activity report format? Are we restricting ourselves to only the groundbreaking news now?

kallisti5:
@pulkomandy If you write an activity report, i'm down publishing it instead.  NVMe is *big* news for most users (one of those sexy changes that grabs headlines)

pulkomandy:
Well I was busy and unmotivated for the march one. Maybe I'll do one in April for the two months.
Maybe we should give it a few days to make sure it got some more internal testing, too?

leavengood:
I am sure they are a fair amount of work, but I love the activity reports (as I am sure most people do), so please do a catch-up one if you can. As I get more involved in Haiku again I could try to help with them too.

I do think the NVMe news should be published separately though. This is good news and sexy, as kallisti5 says. And one more reason for me to build that new Ryzen box with an NVMe disk! 😄

PR #290, opened 2019-04-30
Coding Guidelines: document a redundant "private:"

korli:
to separate the fields from the methods


korli:
This is the follow-up on https://review.haiku-os.org/c/haiku/+/1350

PR #292, opened 2019-05-02
Writing video drivers for BeOS/Haiku 1.2 update

kenmays:
English translation per rudolf's 1.2 doc.


waddlesplash:
Probably we should move these to the tree. But it's OK for now I guess.

PR #295, opened 2019-05-12
update prerequisites for Haiku/pkgman

korli:
It seems mtools is required for EFI on x86_64, python3 for HaikuDepot code generation.


korli:
This would need a better OS icon, but I couldn't find any in the static images folder.

PR #305, opened 2019-09-21
Include missing screenshots into blogpost.

return:
This includes the missing screenshots showing the Chrome debugger console working remotely from macOS to Haiku and another one demonstrating socket.io.


humdingerb:
Thanks!

PR #308, opened 2019-10-01
Fix hugo deprecates

mmuman:
Hugo 0.57 gives some deprecation warnings, this replaces the deprecated constructs.


mmuman:
There's also this which I'm not sure what to do with:
`WARN 2019/10/01 13:12:29 In the next Hugo version (0.58.0) we will change how $home.Pages behaves.`

mmuman:
just rebased.

waddlesplash:
manually merged.

PR #331, opened 2019-12-21
Added guide for running Haiku on a Vultr instance.

R4H33M:

PR #332, opened 2019-12-22
vultr virtualizing article fixes

R4H33M:
It seems like my Vultr article had some errors - #331 
I think I've fixed them now.


humdingerb:
Thanks for the fix! Nobody knows exactly how those markdown pages turn out at the website until they go live. Have another look and don't hesitate to create PRs for further fixes.

PR #336, opened 2019-12-27
guides/building/pre-reqs: Add GCC2 xorriso requirement.

agmsmith:
To compile Haiku under GCC2-hybrid Haiku, you need to manually install the xorriso CD/DVD file system creation tool.  Update the documentation to reflect that.  Also use a more generic package icon, not the Ubuntu one.  Couldn't find a specific Haiku icon!


agmsmith:
Changed to use the more generic "pkgman install cmd:xorriso" as suggested by WaddleSplash rather than "pkgman install xorriso_x86".

waddlesplash:
Probably this should just be added to general requirements?

agmsmith:
Good idea, reworked the commit to add cmd:xorriso to the general prerequisites.

waddlesplash:
Thanks!

PR #359, opened 2020-03-11
Add link to DVD sleeve folding "Burning the Haiku DVD" page

humdingerb:
* replace "CD" with "DVD"
* fix outdated URLs, use https where possible
* link to the sleeve folding instructions
* add the papersleeve as WonderBrush file as a template so people
  may create their own


content/get-haiku/burn-cd.html
```diff
@@ -48,22 +51,22 @@ <h4>OpenBSD</h4>
...
 
 <h4>Windows</h4>
 
 <ul>
  <li>InfraRecorder (open source): <a href="http://infrarecorder.org" title="InfraRecorder website">http://infrarecorder.org</a></li>
- <li>ImgBurn (freeware): <a href="http://www.imgburn.com" title="ImgBurn website">http://www.imgburn.com</a></li>
- <li>ISO Recorder (freeware): <a href="http://isorecorder.alexfeinman.com/isorecorder.htm" title="ISO Recorder home page">http://isorecorder.alexfeinman.com/isorecorder.htm</a></li>
- <li>Active@ ISO Burner (freeware): <a href="http://www.ntfs.com/iso_burner_free.htm" title="Active@ ISO Burner">http://www.ntfs.com/iso_burner_free.htm</a></li>
+ <li>ImgBurn (freeware): <a href="https://www.imgburn.com" title="ImgBurn website">https://www.imgburn.com</a></li>
+ <li>ISO Recorder (freeware): <a href="https://iso-recorder.en.softonic.com/" title="ISO Recorder home page">https://iso-recorder.en.softonic.com/</a></li>
```
pulkomandy: Please use the actual website:
http://isorecorder.alexfeinman.com/#/

softonic is some random software archive and I don't know if they can be trusted or if they add random spyware to what they provide.

content/get-haiku/burn-cd.html
```diff
@@ -1,22 +1,25 @@
...
+<strong>A word of caution on burning DVD images:</strong>
 Some disc burning software may try to be "clever" and won't burn the Anyboot image properly. The following image burning applications are reported to work properly:
 <ul>
 <li>Burn 2.4.1u</li>
 <li>CDRecord</li>
 <li>InfraRecorder</li>
 <li>K3b</li>
 </ul>
-<strong>Using CDBurnerXP 4.3.2.2212 must currently be advised against</strong>, as it reportedly has produced faulty CDs from the ISO images.
+<strong>Using CDBurnerXP 4.3.2.2212 must currently be advised against</strong>, as it reportedly has produced faulty DVDs from the ISO images.
```
pulkomandy: They are at version 4.5.something now. Did they fix the problems? And do we need to list this many tools when just one would work?

content/get-haiku/burn-cd.html
humdingerb: Sorry about that. The original URL didn't work, and I must have messed up when I tried with the domain alone as I got an error. I then just picked a result from a web search...
Anyway, fixed.

content/get-haiku/burn-cd.html
humdingerb: I'm not a Windows user, I don't know about v4.5.
I could reduce the list of Windows apps to "InfraRecorder", which is also listed in the top info box as working.

waddlesplash: Yes, InfraRecorder is the way to go on Windows.

humdingerb: Alright. Only keeping InfraRecorder.

PR #380, opened 2020-06-08
Update MacOS instructions to reflect changes in Catalina

mull:


content/guides/building/pre-reqs.md
```diff
@@ -139,24 +139,26 @@ sudo portinstall devel/bison devel/git devel/nasm lang/gawk print/texinfo sysuti
...
 ## ![osx](/files/os-icons/macosx-32.png) macOS
 
-A case-sensitive file system is required to build Haiku. You can use Disk Utility to create a case-sensitive disk image and store the Haiku source tree on that. Case-sensitive HFS+ works fine.
+A case-sensitive file system is required to build Haiku. You can use Disk Utility to create a case-sensitive disk image and store the Haiku source tree on that. Case-sensitive HFS+ and AFS work fine.
 
 First install Xcode via ```xcode-select --install``` and accept the license. ```xcodebuild -license```
 
 Once you have installed XCode and the command line tools as well as agreed to the end user license you can install the prerequisite software either by using MacPorts or by using Homebrew.
 
+With the release of MacOS Catalina, Apple no longer provides an installation of gcc. The command ```gcc``` is an alias for clang. When building Haiku you need to provide it with instructions to use gcc, which can be installed by both Homebrew and MacPorts. This process likely looks something like ```CC=/usr/local/bin/gcc-9 jam -q -j2 @nightly-anyboot```.
```
waddlesplash, line 148: This should not be necessary, the host tools can be built with Clang, and all other tools use the `crosstools` built using `configure`.

waddlesplash: (If using Clang triggered errors, these should be fixed.)

pulkomandy: If i remember correctly from discussing this on IRC, our Jamfiles end up passing -I- to clang instead of using -iquote. Something is probably wrong with us trying to detect gcc version from clang output, and we think it's a gcc2.

mull: Yeah that's correct. Should I open an issue on trac with the build output I get when clang is used?

waddlesplash: Please do. I'll close this then.

PR #383, opened 2020-06-09
get-haiku/r1beta2: Add rit.edu mirror which is now synced

kallisti5:


kallisti5:
links tested. Working as expected.

PR #387, opened 2020-06-19
Add Windows as build platform using wsl2

yann64:
Building Haiku on Windows using Windows Subsystem for Linux version 2 is possible.
Add Windows as an indirectly supported platform through wls2.
Build performance on wsl2 is now close to building from Haiku in a VM (at least on my machine), so removed note of slow performance using wsl1
Corrected microsoft link to use https.

### DIFF
--- a/content/guides/building/pre-reqs.md
+++ b/content/guides/building/pre-reqs.md
@@ -36,7 +36,7 @@ Note that in addition to all the platform-specific packages, you will also need 
 | BeOS                               | [pkg](#beos_zeta)    | NO        | Once upon a time... |
 | macOS                              | [Homebrew](#osx)     | NO        | Need a working case sensitive filesystem |
 | Solaris                            | [solaris](#solaris)  | NO        | No longer supported |
-| [Windows](http://microsoft.com)    | [see notes](#windows)| NO        |                     |
+| [Windows](https://microsoft.com/)    | [see notes](#windows)| USING WSL | [Using Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/) |
 | Zeta                               | [pkg](#beos_zeta)    | NO        | Once upon a time..  |
 
 <a name="pkgman"></a>
@@ -180,8 +180,7 @@ You can now retry the <code>port install...</code> command in a new Terminal.
 
 It is possible to install Windows Subsystem for Linux and follow the instructions
 for the appropriate Linux distribution (for example, Ubuntu is available this way
-in the Windows store). However, this is currently slower than running Haiku in
-a virtual machine and building from there.
+in the Windows store).
 
 <a name="beos_zeta"></a>
 ## ![beos](/files/os-icons/beos-32.png) BeOS & Zeta


waddlesplash:
This may work (I also use it) but it is unreliable and I do not think we want to call it a "supported" platform.

pulkomandy:
What's unreliable about it? Anything besides lack of filesystem attributes?

nielx:
I have been building Haiku on WSL with success for a while now. And I don't even think WSL 2 is required; I have been using WSL 1.

yann64:
> What's unreliable about it? Anything besides lack of filesystem attributes?

I don't know tbh, tried on the last haiku revisions without any issues,so looks pretty reliable from my experience.

content/guides/building/pre-reqs.md
```diff
@@ -180,8 +180,7 @@ You can now retry the <code>port install...</code> command in a new Terminal.
```
pulkomandy, line 180: The alert saying "not a maintained development environment" can be removed as well, I think?

content/guides/building/pre-reqs.md
```diff
@@ -36,7 +36,7 @@ Note that in addition to all the platform-specific packages, you will also need
 | BeOS                               | [pkg](#beos_zeta)    | NO        | Once upon a time... |
 | macOS                              | [Homebrew](#osx)     | NO        | Need a working case sensitive filesystem |
 | Solaris                            | [solaris](#solaris)  | NO        | No longer supported |
-| [Windows](http://microsoft.com)    | [see notes](#windows)| NO        |                     |
+| [Windows](https://microsoft.com/)    | [see notes](#windows)| USING WSL | [Using Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/) |
```
pulkomandy, line 39: "Using WSL" does not need this much uppercase :)

content/guides/building/pre-reqs.md
```diff
@@ -36,7 +36,7 @@ Note that in addition to all the platform-specific packages, you will also need
 | BeOS                               | [pkg](#beos_zeta)    | NO        | Once upon a time... |
 | macOS                              | [Homebrew](#osx)     | NO        | Need a working case sensitive filesystem |
```
pulkomandy, line 37: macOS should be marked YES again, since we fixed the build problems there.

waddlesplash:
WSL1 does not support extended attributes so you are left with the attributes emulation system. WSL2 supports it I believe, but that does not need a comment here, as WSL2 is just a Linux VM, yes?

yann64:
> WSL2 supports it I believe, but that does not need a comment here, as WSL2 is just a Linux VM, yes?
Yes, uses Hyper-V under the hood from what I understand.

waddlesplash:
Then I don't get the point of this PR, of course building on Linux VMs is supported.

pulkomandy:
It makes it clear that's the recommended way to build things from Windows.

waddlesplash:
But it is not, you can run a VM of anything you like on Windows (including Haiku, which is probably preferred to Linux at this point, if you can pick it?) If you want to develop in WSL, that's fine, but it is not really prioritized over any other VM here technically.

nielx:
WSL really is different in the sense that it is a first class citizen on Windows. You can install Ubuntu from the Microsoft Store! This is what makes it different from using another intermediary virtualisation where you will have to set up an individual container and install/run the linux distribution of your choice. As such, I would be in favour of documenting this explicitly.

yann64:
Exactly this:
> WSL really is different in the sense that it is a first class citizen on Windows.

I use Visual Studio Code and with the correctly configures tasks.json never "directly see" the Ubuntu WSL instance. And sourcefiles and generated binaries are accessible from the windows explorer.

PR #388, opened 2020-06-29
Refactor getting involved info & make it more visible

bitigchi:
Currently "Getting Involved" section is deep in "[Community](https://www.haiku-os.org/community/)" section. The header part opens another page with more technical info, and it's confusing.

There is a [link](https://www.haiku-os.org/community/getting-involved) on the main page, however that page opens the technical page mentioned above.

This PR adds a new navigation bar item named "Contribute", merges the "Getting Involved" categories table and the more technical and wordy page in one.

It is not tested, I don't know how to build the web page, so I'd be glad if anyone can point that out to me. If not possible, please review thoroughly, as it's my first significant attempt to modify the website. Any ideas for improvement are welcome.


pulkomandy:
I would prefer if we kept the files where they were (community/getting-involved) to avoid creating deadlinks (from outside the website) if possible (the URLs don't have to reflect the website organization).

Don't know about adding it to the navigation bar. I mean, it's "just one more thing" there but we could also add "getting support" and many others there. I would prefer a larger rethinking of the navbar, as I never was really happy with the current categories (they are fine for organizing our pages, but not great for users who navigate the website). In particular the "documents" page in its current state is useless (has been since 2008 or so), and should be replaced with /guides/, and the technical docs that used to be in documents/ be linked from somewhere in the development/ pages (or guides/, if they are user-oriented).

Well, I guess the extra item is ok as a bandaid but not in the long term, as we're just going to add everything to the navbar if we go this way...

waddlesplash:
> I don't know how to build the web page

Run `hugo serve` in the repository root, this will start a temporary webserver with the built site.

waddlesplash:
We may want to reorganize the links or change the page texts, but I don't think there's anything wrong with the old layout per se.

bitigchi:
The idea is to make contributing info visible with one click. Current getting involved link and the page is not very helpful, too much babble. The table with icons in the Community page is okay, but I am not sure if the Community label is understandable enough.

Ideally there should be a direct Contribute/get involved link from the main page with related sole content.

waddlesplash:
Yes, then changes should be made to the page contents but we do not need to move things around here.

PR #396, opened 2020-07-17
about/archival: Add a page mentioning archival locations for Haiku

kallisti5:
* We may want to extend this to include information about backups?


waddlesplash:
I'm not sure what the value of this page is...

kallisti5:
> Will this page be linked from anywhere?

Should show up as a subpage of the about section.

> I'm not sure what the value of this page is...

Listing all of our repositories and everything we're doing to protect our source code seems like a good thing to document... might as well advertise all the places we back it up as well.

It maybe something known to a few of us, but why not let everyone know?   It's pretty impressive how many places our VCS is replicated to in how many forms.

content/about/archival.md
```diff
@@ -0,0 +1,41 @@
...
+  * Developers all over the world have local copies of the Haiku repositories on their desktops/laptops/workstations
+
+## Cold Repositories
+
+> These are source code repositories which receive occasional updates, and are generally more complex to restore.
+
+* [Software Heritage Archive](https://archive.softwareheritage.org/browse/search/?q=https%3A%2F%2Fgithub.com%2Fhaiku%2F)
+  * This group accepts requests to periodically archive git repositories for long-term archival.
+* [Github Artic Vault](https://archiveprogram.github.com/)
+  * In 2019, Github began a yearly program to archive public repositories in an Artic Vault on [piql](https://www.piql.com) film.
```
return, line 40: Small typo here. Artic -> Arctic

pulkomandy:
I think in the current state this just results in free advertising for github. We just have to know for ourselves that this exists and hope we never need it.

jt15s:
Perhaps instead of a page on the website, we could include this as an MD file within the source code for Haiku?

leavengood:
I am working on these stale PRs. I would like the PR list to be empty. Do we want this page or not? I feel like it doesn't really hurt anything. I see some typos though so I would want to fix those first, but I can do it here in GitHub.

content/about/archival.md
```diff
@@ -0,0 +1,41 @@
++++
+type = "article"
+title = "Archival"
+date = "2009-09-09T21:27:31.000Z"
```
leavengood, line 4:
```suggestion
date = "2020-07-17T13:27:31.000Z"
```

content/about/archival.md
```diff
@@ -0,0 +1,41 @@
...
+tags = ["security", "archival", "backups"]
++++
+
+# Source Code Archival
+
+Haiku's source code represents hundreds of thousands of hours of work by individuals across the globe.
+To protect their investment, Haiku reguraly ensures multiple copies of these critical source code
```
leavengood, line 11:
```suggestion
To protect their investment, Haiku regularly ensures multiple copies of these critical source code
```

leavengood (commented):
I would fix the date and there is also this typo.

Coldfirex:
I think there is benefit to this page.  At the very least its a reference point for maintainers, etc. on a bit of our infrastructure which we could expand on more in detail down the line.

PR #398, opened 2020-07-27
Add screen shot of a KDL to Welcome to Debugging Land.

agmsmith:
After that recent discussion https://dev.haiku-os.org/ticket/13321 about blue screens of death, and addition of Haiku's equivalent to Wikipedia, I realised we didn't have a screen shot of a KDL for comparison.  So I made a typical screen shot and added it to the documentation about Kernel Debugging Land.  Now we'll have something to cite for Wikipedia.  Hope I got the paths right in the URLs for the image.


mmuman:
Else don't forget to run kdlhangman if you make a new one :-)

agmsmith:
On 2020-07-27 11:20 a.m., Alex von Gluck IV wrote:
 > @agmsmith do I have your permission to use this screenshot on the KDL 
wikipedia page? :-)

Yes, you may use it on WikiPedia.  That screen shot is for public 
viewing, so consider it released to the public domain.

On 2020-07-27 12:20 p.m., François Revol wrote:
 > Else don't forget to run kdlhangman if you make a new one :-)

Hmmm, hangman screen shot?  How about that QR code plugin for stack 
traces?  Anyway, I just wanted the typical stack crawl and general 
screen overwritten look.

- Alex

PR #399, opened 2020-07-30
Ticket-16415 Add OpenBeOS Newsletters to site

q5sys:
PR to fullfill: https://dev.haiku-os.org/ticket/16415


X547:
Ghostery leftover:

`<script src="" type="text/javascript"></script><style>@media print {#ghostery-purple-box {display:none !important}}</style>`.

q5sys:
> Ghostery leftover:
> 
> `<script src="" type="text/javascript"></script><style>@media print {#ghostery-purple-box {display:none !important}}</style>`.

good catch I'll do another pass and clean any of those out.

waddlesplash:
The inline data: for CSS and images is really ugly... can the original documents and files not be retrieved?

q5sys:
> The inline data: for CSS and images is really ugly... can the original documents and files not be retrieved?

For images over a certain size archive.org saved them as base64 data. To decrease the complexity of their hosting and storage costs, archive has taken some shortcuts to provide their service.  They do not simply do a scrape of a site and save the assets as is.  A long long time ago they never even captured images because of how fast their storage needs would have increased if they did.  They came to their current system as a sort of best case situation for archiving sites.  They never intended to be a bit for bit replica, rather to save the important stuff for later access.

Hypothetically the base64 encoded data could be converted back to images, but that would be a manual process, I'm not aware of anything that will do that automatically that's free or anything that's open source to automate that process.

pulkomandy:
Let's merge this, and clean it up later. We may extract the contents and integrate it in the current website engine/theme, for example.

PR #402, opened 2020-08-11
Date fixes and new blogpost

CruxBox:


pulkomandy:
It looks like you are re-adding your previous post? Why is that?

CruxBox:
> It looks like you are re-adding your previous post? Why is that?

Hey, the previous post had the date wrong. So I changed it.

pulkomandy:
You added a second copy of it.  think it's better to fix the date in the header of the file, but not rename the file with the new date. That's because the filename is visible in the url, and moving that will create deadlinks (in the forum thread with comments about the post, for example)

CruxBox:
I messed up my local a bit. I'll create a new PR.

PR #417, opened 2020-09-17
Clean up getting involved pages and update stuff

bitigchi:
Thanks @pulkomandy for the guidance. Sorry for the commit history mess.


pulkomandy:
The update is welcome but I think we should move this list of projects to Trac tickets, which would enable people to vote on them, and developers to better keep track of what is and isn't done. Ok to merge for me but let's think about that :)

bitigchi:
For projects like Web+, unfortunately the number of issues makes it impractical to list all of them. I will try to list missing features, and then provide a link to all Web+ issues in Trac.

pulkomandy:
I didn't mean just linking to existing issues (we don't want to flood unsuspecting newcomers with a detailed list of bugs here). I meant move the text that's here on the website to Trac tickets (enhancement tickets or we can create a new ticket type if it's more convenient for that). This would preserve the high-level view we have here (which is good), but allow to have ticket dependencies, milestones assignment, and voting.

bitigchi:
That would be cool, but in the meantime we can merge this until someone edits the tickets with updated descriptions and blocking-blocked by's.

PR #421, opened 2020-09-24
Specify PCM format

bitigchi:
As pointed out in the comments, it's better to just specify PCM format, instead of just WAV, which can serve as a container for other compressed formats.


humdingerb:
Thanks

PR #422, opened 2020-09-27
Sound contest: clarification and re-ordering

humdingerb:
* Added clarifications where the sound description may not be clear enough.
* Renamed 'Error' category to 'General' and moved the start chime there.
  The 'Beep' isn't necessarily a sign of error, it can be used for various things.
* Re-ordered the notifications according to their severity.


humdingerb:
As AFAIK the notification sound for "progress" isn't implemented yet, did I get the description right? Is the sound played whenever the progress bar is updated?

PR #428, opened 2020-10-01
Improve docs and grammar

Hemraj0007:


pulkomandy:
Sorry, but you are not improving anything here. The grammar was correct.

In any case, changing two words in a webpage is not going to be of much use. Try to work on something a bit more substantial.

PR #434, opened 2020-10-02
Add contest entry for Daniel Diaz

bitigchi:


humdingerb:
Shouldn't the contest entries be at the end? People happening upon the news post may be taken aback seeing the entries before the intorductory paragraph and details of the contest. And people knowing about the contest should have no problem checking at the end of the article for the latest additions.

Also, you should check into "feature branches" when doing PRs. That'll make sure you don't have remnants of past PRs in your current one.

humdingerb:
I expect this and other entries will receive updates by their contributers, based on feedback etc.
I think, we may want to add a "(v1.0)" to the entry, and also re-package the zip so the folder also says "...(v1.0)".
Then increase the version for every update.

bitigchi:
> I expect this and other entries will receive updates by their contributers, based on feedback etc.
> I think, we may want to add a "(v1.0)" to the entry, and also re-package the zip so the folder also says "...(v1.0)".
> Then increase the version for every update.

Instead of version numbers, we can just do dates, like last updated. Putting version numbers to other people's work seem weird to me.

humdingerb:
> Instead of version numbers, we can just do dates, like last updated. Putting version numbers to other people's work seem weird to me.

Also possible. Maybe a "(submitted on 2020-10-03)" and then "(updated on 2020-10-08)" after the download link?

PR #435, opened 2020-10-03
Update README.md

prashant-45:


humdingerb:
Nope. The section is called "Troubleshooting", "section" stays lower case.

humdingerb:
I wanted to set the label to "invalid", but neither WebPositive nor Otter let me...

pulkomandy:
Done. Also, after much complaints about this, Hacktoberfest have changed their rules so that only repositories with the "hacktoberfest" topic are taken into account for their program, hopefully reducing spam in other places. So the label should not be needed anymore for now.

PR #441, opened 2020-10-16
Update contact.html

nagpalnipun22:
Add the body HTML and title tags
Added the head section


pulkomandy:
This is not needed. We use the Hugo website generator and it takes care of adding these tags for us already when the website is put online. As you can see if you visit the website at www.haiku-os.org, you can confirm there that the body and head tags are indeed present.

PR #442, opened 2020-10-16
Add entries for Humdinger and Julius Enriquez

bitigchi:


humdingerb:
You've linked to the "Tumble Leaves" zip, but uploaded the PNG.

Should a preview PNG be added? Maybe use 1 1920x1080 as preview, but force the displayed size in the post to 1/4 (480x270)?

bitigchi:
We can add previews, but entries like Julius', which includes many images might be a problem. It has 12 images, I guess we'll have to indulge ourselves on which to include.

humdingerb:
As long as they are of a similar theme, maybe just show one and state that this is just "one example of the x submitted backgrounds".

Don't you want to always link to the zip archive? It contains the workfile after all...

bitigchi:
To the post? Post should only contain the previews, not the working files. We need to find another place to upload the working files. I made a custom `.zip` for the Julius' preview files.

humdingerb:
Oh, alright. I do a preview pic to every entry should be added. If there are more, just pick one and mention that the download link is a zip with a preview of all x submitted backgrounds.

bitigchi:
How do you resize the preview display?

humdingerb:
Should work with  I suppose. Not sure though how well that'll look. I expect there's no filtering... At  least in the forums, the website shows a smaller preview automatically. Maybe see first, if the rest of the website does too... :)

humdingerb:
Oops. Should have been `< img src="xxx" width="xxx"height="xxx" />`

PR #446, opened 2020-10-19
Updated r1-final-features-poll.html

Nishthajosh:
target=”_blank” help to open the link in new tab....
Hope this will helpful to you.


humdingerb:
Making such a change of UX on a single page of the website makes no sense.
More importantly, we want to leave it to the user if a page opens in a new window/tab or in the current one. People can middle-click on links to open in a new tab.

PR #449, opened 2020-11-20
Update to policies to cover more aspects of the project.

aplgithub:
As discussed recently; these changes are to;

* Tidy up and clarify the language around the previous content
* Clarify the position of community members / volunteers
* Clarify the release intent

This is totally open to discussion and amendment and the first commit represents a starting point.

The idea is that after this is settled, it will be possible to point people in difficult situations to this document for clarification.


waddlesplash:
I don't know, most of these additions seem too verbose, or altogether superfluous and purely reactionary to recent problems to me, and the sections that @pulkomandy suggested we really should expand on (most notably, the ones about what to do when people are behaving inappropriately) seem to just be the same as before. I especially am not a huge fan of the "glossary", it makes this document feel "legal", which it really is not.

I feel like most of the additions about "project members are not obligated..." would be much better as a 1-sentence aside, rather than something that sounds like a contract disclaimer: the fact of the matter is, the developers *do* have some kind of obligation to the project; the question is the why, the what, and the how of that.

If the major motivation for doing this is mostly so we can have something to point to when someone is causing problems as a justification for disciplining them, then we do not really need to lay all this out in such fine detail (or, indeed, should not in the first place); broad strokes are more than sufficient.

content/community/organization/policies.md
```diff
@@ -0,0 +1,110 @@
...
+
+* where security or privacy concerns need to be raised and where the information should not be public
+* where another community member's behavior is inappropriate or offensive and you need to address the issue
+* where contributors are collaborating about specific details and both parties are happy to communicate in this way
+
+### Inappropriate Content and Behaviour
+
+Discussions relating to subjects that are clearly off-topic such as religion, politics, gender and ethnicity should not be undertaken across any of the communications channels.  Community members should also refrain from making sexually suggestive comments or using sexual inuendo toward other community members using any of the communications channels.
```
pulkomandy, line 42: The previous wording was, I think purposefully, more tolerant than that on off-topic subjects. With this version we would be disallowing any discussion, for example on gender and ethnicity, which would prevent discussing things like "how to improve diversity in open source software".

It is fine to have discussions going off-topic occasionally. We love to get news from Begasus' dogs on the IRC channel, for example. But sometimes it gets out of control.

So I would prefer to keep the previous wording of "refrain from delving into"?

content/community/organization/policies.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Glossary
+
+|Term|Meaning|
+|---|---|
+|Community Member|The term "community member" is used in this document to refer to any individual who is engaged with the Haiku project in any way.|
+|Communications Channel|The term "communications channel" is intended to cover all communications channels related to the Haiku project.  Examples of "communications channels" include IRC, mailing lists, emails, code-reviews, discussion boards and other media.|
+|Development Team|The term "development team" is used to refer to a group of individuals who have commit access into the Haiku source code repository and who are recently actively engaged on the project contributing to the source code.|
+
+## Volunteers
```
pulkomandy, line 18: I would rather have this and the definition of "development team" moved into the "governance" page.

content/community/organization/policies.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Nightly Images
+
+Haiku's nightly images are a build of the entire operating system that contains the latest enhancements, bug fixes, and hardware support. Please keep in mind that there may also be regressions and broken features in these builds which have not yet been identified or resolved. Users should only use a nightly build at their own risk and be aware of the potential for experiencing issues.
+
+### Formal Releases
+
+The Haiku project aims to make an annual release of the operating system.  In general a formal release will tend to be made in the northern hemisphere spring / southern hemisphere autumn.  The rationale for this timing is because it is the time of year at which the development team are most likely to have capacity to deal with a release.  At the release time, the development team will attempt to stabilize the state of the project and to create a reference point for users of the operating system.  A release may be deferred or may be cancelled entirely if there are difficulties making the release for any reason.
```
pulkomandy, line 92: I don't think this should go in the policies either. As I mentionned on the mailing list, some things are unwritten because they allow us to change our mind on how we do things. I think we should move this to a separate page and probably put it in the Trac Wiki to emphasize that it is editable by any contributor.

content/community/organization/policies.md
```diff
@@ -0,0 +1,110 @@
...
+
+Some community members are occasionally frustrated that a feature or function that they would like is not available on the platform or is not sufficiently developed for their actual or perceived purpose.  In such situations, keep in mind that the contributors to the project are doing so mostly on a volunteer basis and because the work interests them.  Although often times, contributors are interested to hear about any problems that users are having, the contributors are not beholden to users' requirements.
+
+If you are unable to contribute materially to the software yourself, the best way that you can achieve the goal of having a feature implemented is by using the [bug tracker](https://dev.haiku-os.org) system.  Identify if there is an existing ticket that correlates with your requirements or if there is no existing ticket then create one.  You should document factually, clearly and accurately the problem you have.  Make sure there is enough detail that the problem can easily be understood and reproduced by a contributor.  If there are obvious resources that might help such as input / output files, screenshots, logs etc... then attach those to the ticket directly.
```
pulkomandy, line 106: The bugtracker already has documentation on how to write a good bug report and there is no need to repeat it here. IT kind of loses the main point, which is that support is provided by volunteers in a best-effort way only, and it can be frustrating because it causes large delays in handling of tickets.

Other things could be clarified however:
- The possibility to vote on existing tickets to raise the attention of the devs,
- The bountysource page if people want to put money on a specific ticket,
- The fact that donations to Haiku inc are only donations and do not mean we owe them anything. In particular, the Inc and the Haiku project are independant entities and there are alternatives (the HSA, direct donations to developers on Liberapay, ...)

pulkomandy (commented):
I think the additions are useful, but I would organize things a bit differently (to keep this page short) and the tone is more formal than we usually do (don't know if this is a good or a bad thing).

aplgithub:
Thanks for the feedback; unfortunately I've had some stuff come up and can't deal with this right now -- hope to get back to it in a week or so.

pulkomandy:
Following up on even more discussions on this subject on the forums and IRC, I have made a different proposal which keeps the original language but adds some clarifications and extra limitations based on past events and plausible future ones: https://github.com/haiku/website/pull/460

aplgithub:
Hi @pulkomandy - unfortunately my time is limited right now to engage.  I will close this PR and switch over to your one.

PR #453, opened 2020-11-26
Fixes

bitigchi:
Fixes mentioned by Starcrasher on forums.


bitigchi:
I'm confused. I replaced Tim L's entry, but it only shows the deletion within the changes. Did I do something wrong?

humdingerb:
Not sure what's going on. Did you change something with Tim's zip?
Anyway, it looks like the archive will still be intact after this change. I merge and we'll see...

BTW, it's better to have a regular commit 'title' and put the "Fixes #453" and the end of the body of the commit message.

The deadline for the contest has past yesterday, which I wasn't aware, actually. :) Should we put an info box at the top to have people stop submitting new entries?
Also, did you have a think how to do the poll which themes to include? Just a normal poll where people check the ones to include, I suppose. But how to process the results? Just cut off all entries below a certain gap in votes? Like if the 5th placed has 50 votes and the 6th only 12, all below the 5th didn't make it? A bit subjective and arbitrary...
Any better ideas?

bitigchi:
I pulled changes to my master branch, it looks like the file is intact.

I commented on the comments thread, I think that should be sufficient. It is indicated within the text anyway.

I was thinking about listing all the entries in the poll, and making it single selection; so people vote which one they like the best. Hopefully this will lead people to vote more selectively. First vote gets to be the default, other ones are either additional choices included by default or packaged for optional install.

Am I missing something?

humdingerb:
I'd find it very hard to pick just one single theme. OTOH, it's quite possible everyone picks the same 5 top themes plus a few different ones. Maybe that's not a bad thing, and we'll have a second, single-vote poll to determine the actual default, with the other themes as additional/optional installs.

humdingerb:
You think we should involve the general mailinglist to see if there's experience/advice on these kinds of polls?

bitigchi:
> You think we should involve the general mailinglist to see if there's experience/advice on these kinds of polls?

Maybe we can ask the #haiku-dev channel first. It sounds too trivial to ask on a mailing list.

pulkomandy:
The usual thing we do (for example to rank GCI winners) is Condorcet / Borda count (https://en.wikipedia.org/wiki/Borda_count).

Debian uses a similar system (it happens there: https://surveys.debian.net but the survey for the next theme closed a few weeks ago and now there's nothing to see in that website). It tends to select the less controversial option and gives a bit more nuance to the votes, while not being too complex (voters just have to rank their preferred entries in order)

bitigchi:
> The usual thing we do (for example to rank GCI winners) is Condorcet / Borda count

Neat. Do you do the calculations manually or is there a free service we can use? If I'm not wrong, Discuss cannot do this.

pulkomandy:
For GCI voting we used Google Forms and Google Spreadsheet to manage the results. I think it's the easiest thing to set up.
If we plan to do this often we could consider setting up our own limesurvey instance like Debian did (the free tier on LimeSurvey cloud offers seems too limited for our needs, so self-hosting would be required).

bitigchi:
> For GCI voting we used Google Forms and Google Spreadsheet to manage the results. I think it's the easiest thing to set up.

I am now looking at Google Forms, and the interface is confusing as hell. Do you have an archived link of a similar poll done before (or any guidance for that matter)?

> If we plan to do this often we could consider setting up our own limesurvey instance like Debian did (the free tier on LimeSurvey cloud offers seems too limited for our needs, so self-hosting would be required).

For the wallpapers we can do something similar.

pulkomandy:
We had a single "multiple choice grid" with rows for "first", "second", "third", etc and a column for each candidate (each theme, in this case). "limit to one response per column" enabled.

Here is one I set up for the sound theme with these parameters: https://forms.gle/eK8p9XYhfcVnP2jf8

Limesurvey has a probably easier approach to this, where you just drag and drop the options to rank them.

bitigchi:
This is my take: https://forms.gle/6ejCHVGZDvQhnJg16

I think rows work better in this case, since we have many contestants.

pulkomandy:
Yes. It seems you didn't enable "single result per column" however (in the "..." menu at the bottom right of the form editor). So it's currently possible to rank everyone as first.

bitigchi:
Fixed that: https://forms.gle/svbpbfT27kBNhakZ9

It looks like there is no way to auto-close it at a certain time.

humdingerb:
Good work!
So the poll is ready to go? How does it deal with making sure people only get to vote once? via IP address? I suppose voters don't need to be logged into a Google account.

bitigchi:
We can require a Google account to make sure they only vote once, but I'm not sure if that's too friendly. We can do that if there is consensus though.

I'll open a thread tomorrow at these times (coincides with New Zealand), and we'll make it live.

humdingerb:
That's definitely a downside to Google forms. Using it to get the votes from GCI mentors is unproblematic. With the weirdos on the net (I remind you of the alleged satanist, if you caught the thread in the forums before I removed those posts), ballot stuffing is a real threat. I guess we can always book Rudy for a gig at the 4 Seasons, if the results are inconvenient... :D

bitigchi:
I think that the community will make sensible choices though, I don't think anyone would vote for some disturbing theme that they'd constantly hear all the time. I got my choices ready!

And hopefully someone will step up and fix the sound issues and make it more friendly to select different themes :D. But I'm sure eventually it will all be fixed, not in a rush anyway.

PR #454, opened 2020-11-28
Fix version with missing beep

bitigchi:
This one was updated before the deadline with the missing beep sound, but apparently I mistakenly included the other version.


bitigchi:
Nope, false alarm!

PR #464, opened 2020-12-27
Show all the posts from a blog in its list

lonemadmax:
Currently /blog/\<blogname\>/ is somewhat of a mix of things. It shows all the site posts with the same author as that in /blog/\<blogname\>/_index.md. That means:
1. You only get the list for created blogs. No articles from people without their blog section.
2. (Consequence of 1) You don't get articles with variation in the author property (pawel_dziepak vs paweł_dziepak, pulkomandy vs PulkoMandy, etc)
3. You get posts from all sections, not only the blog one.

Given the URL prefix, 1 seems reasonable but 2 and 3 not so much. Now, if /blog/\<blogname\>/ is to be seen as an author and not a blog entry point, then 3 is a feature and this change can be rejected on the spot, preferring other solutions.

### DIFF
--- a/themes/shijin4/layouts/blog/list.html
+++ b/themes/shijin4/layouts/blog/list.html
@@ -9,7 +9,7 @@
 
 		<div class="col-md-9">
 		{{- if (isset .Params "author") -}}
-			{{ $paginator := .Paginate (where .Site.RegularPages ".Params.author" "==" .Params.author) }}
+			{{ $paginator := .Paginate .RegularPages }}
 
 			{{- range $paginator.Pages -}}
 				{{ .Render "summary" }}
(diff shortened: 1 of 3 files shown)


waddlesplash:
I think we should list all pages in the given blog directory, yes. Is that what this change does?

lonemadmax:
Yes. Instead of going through all the site pages with a given author, it just retrieves the pages of the current context, which for /blog/\<blogname\> is a specific blog directory.

Two easy test cases where the difference is easy to notice are https://www.haiku-os.org/blog/suhelmehta (one of the articles has different author capitalization and is not currently shown) and https://www.haiku-os.org/blog/digib0y/ (two different "ortographies", none of which is the Deepanshu that is in the _index.md file).

There's still an issue I hadn't noticed before (not in the list, but you may prefer the solution to that also be included in this PR). The header in the full articles with a different author property link to a page for a non-existent list, as they go for author instead of blog.

pulkomandy:
We should also fix these mismatched author tags in blogposts.

PR #469, opened 2021-01-21
News: Remove duplicate sound set entry

CodeforEvolution:
The "System Sounds Contest Wrap-up" news article had a duplicate entry in the results table.


humdingerb:
Thanks!

PR #471, opened 2021-01-24
Automatic year update

lonemadmax:
There's still the copyright line in config.toml, that can't be programmatic, AFAIK.
For the donation chart I'm assuming "calendar years" are used, and not May 4th 2020 - May 3rd 2021 or whatever. The financial reports seem to agree, but someone please confirm before applying change.


leavengood:
Yes the donation meter year should be just based on the current year, pulled from the file like you did. This is a good change, thanks. In fact I wanted to do something similar.

leavengood:
Thanks again, this made it much easier for me to get the donation meter updated for 2021. 👍

PR #472, opened 2021-01-27
shijin4: Fix dark mode to be a bit more refined

kallisti5:
Cleanup our dark theme to be a bit better.

* #000 background to more gray
* Fix "bright" objects to be muted
* Fix misc UI elements which were washed out.

# BEFORE
before-3
before-2
before-1

# AFTER

after-1
after-2
after-3

### DIFF
--- a/themes/shijin4/static/css/shijin4.css
+++ b/themes/shijin4/static/css/shijin4.css
@@ -513,7 +513,7 @@ ul.pagination {
 	border-color: #c7c7c7;
 	border-style: solid;
 	border-width: 1px;
-	padding: 0px 2px 0px 2px;
+	padding: 3px 2px 0px 2px;
 	background-color: #e8e8e8;
 	font-family: serif;
 	font-variant: small-caps;
@@ -529,6 +529,8 @@ ul.pagination {
 	background-color: #e8e8e8;
 	font-size: 0.9em;
 	font-family: monospace;
+	padding: 2px;
+	border-radius: 3px;
 }
 
 figure {	/* Use for step-by-step guides consisting mainly of */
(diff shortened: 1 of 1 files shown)


kallisti5:
Don't take this as perfect.. but just an improvement of the "worst offenders"

kallisti5:
We might also want to change the "static image" background gradient to a css gradient...

{{{
Light "normal":
background: linear-gradient(#121933, #3b5378) fixed 0% 0%;

Dark:
background: linear-gradient(#111, #222) fixed 0% 0%;
}}}

kallisti5:
for nephele who wants ```#000```

Dark mode sites not  ```#000```...

* github ```#0d1117```
* duckduckgo ```#1c1c1c```
* discourse ```#111``` (try.discourse.com)


**https://99designs.com/blog/web-digital/dark-mode/**

> 1. Don’t go too dark
> 
> Book publishers don’t use pure white paper because the contrast with black ink is too stark. It makes you squint and can lead to headaches. The same goes for digital devices. Steer clear of pure black. It’s hard on the eyes to look at a high contrast screen. Good dark mode colors are shades of grey combined with desaturated colors.


**https://css-tricks.com/a-complete-guide-to-dark-mode-on-the-web/#design**

> Dark Mode Colors
>
> Pure white text on a pure black background will look jarring. The trick here is to use an off-white for the text and off-black for the background. Material Design Guidelines for example recommends #121212 for the background.

kallisti5:
as this represents large improvements to dark mode (making it more usable) and I haven't heard any push back to it beyond changing to background to "not 100% black" (which i provided documentation on why it's a bad idea)... i'm merging it.

PR #473, opened 2021-01-28
Update Haiku Slideshow

bitigchi:
Updates the screenshots with the current state of the system.

Also adds a ported apps screenshot, with a screenshot including
LibreOffice, Krita, and Konsole.

[screenshots.zip](https://github.com/haiku/website/files/5889708/screenshots.zip)


humdingerb:
We should seize this opportunity to move all screenshots into a "slideshow" subfolder.
I woud have stayed with 16:9 dimensions, but I suppose that's personal taste...

bitigchi:
There is already a folder named slideshow, but I didn't know what to do with the images inside that folder. I can still move them though.

Regarding the aspect ratio, I also like wide screen better. But I don't have any installation with such aspect ratio, couldn't get VBox to work with it.

humdingerb:
Indeed, there is a "slideshow" already. With another folder "Haiku_Slideshow" in it... I guess that's a very old version of a screenshot slideshow. Let's keep that for now and just move the images of our slideshow into "slideshow".

It's been a while since I used VirtualBox, but the tip on setting screen resolution did work for me for 16:9.
https://www.haiku-os.org/guides/virtualizing/virtualbox#part_customVESA

bitigchi:
Moved the screenshots to slideshow folder.

For some reason above video mode thing does not work on my computer. Tried it before, but no luck. Anyway, now that I think about it, 4:3 screenshots are more accessible, and offers more data at first sight.

humdingerb:
For actual work on a desktop computer I'd prefer 4:3 as well, or at least 16:10. But as a screenshot tour, with most people probably in front of a 16:9 screen, a 4:3 slideshow isn't ideal. 16:9 previews on the page also look nicer and lead to less scrolling.

You could of course, set a background image with a horizontal line at the bottom edge of a 16:9 resolution. Then make sure everything is above that line when you take your screenshot and later batch crop everything to 16:9 (with e.g. GraphicsMagick). But I understand that you'd not be thrilled to do it all over again...:)

bitigchi:
As I indicated above, I am not able to make 16:9 screenshots. If anyone would like to change the aspect ratio, feel free to take over.

humdingerb:
As I submitted, screenshots can be cropped from 4:3 to 16:9. I may look into it soon...

PR #487, opened 2021-04-01
Change footer to qualify for Netlify open source plan

leavengood:
I believe these changes will make Haiku satisfy the [Netlify Open Source Plan Policy](https://www.netlify.com/legal/open-source-policy). I decided to use one of their [badges](https://www.netlify.com/press/#badges) but we could also qualify by linking to their homepage with a text link that says "This site is powered by Netlify".

I added a link to the Community guidelines to satisfy their second bullet point in the above policy.

We were on a free plan for a long time but upgraded our plan in June last year because of more traffic due to the Beta 2 release. We have been paying quite a bit each month since then. I would like to fix that with this.

Once this is merged, we can submit their form to try to get Haiku on this plan.

I am open to adjustments to this.

This is what this looks like on my windows machine:

footer

### DIFF
--- a/themes/shijin4/layouts/partials/footer.copyright.html
+++ b/themes/shijin4/layouts/partials/footer.copyright.html
@@ -1,4 +1,14 @@
 <footer>
+	<div style="float: right">
+		<a href="https://www.netlify.com">
+			<img src="https://www.netlify.com/img/global/badges/netlify-dark.svg" alt="Deploys by Netlify" />
+		</a>
+	</div>
+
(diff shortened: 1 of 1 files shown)


leavengood:
Here is the Netlify preview deployment of this branch: https://60653695791fe700073f3e38--haiku.netlify.app/

We should probably use this more often and maybe set up a GitHub integration or something.

pulkomandy:
Looks good to me. And yes, we should have some way to preview website changes.

jt15s:
Agreed - it would be nice to have something like Netlify to do test deployments so any changes to the site can be fully tested.

PR #488, opened 2021-04-02
Blacklist -> Blocklist

humdingerb:
hrev55022 changed the term "blacklist" to "blocklist".
This PR reflects this change in all guides, the faq and the current
2021 GSoC ideas page. Older blog posts remain untouched as re-writing
history (and other people's texts) doesn't feel right (or warrented).

The file names may still contain "blacklist" in order to keep external
links working.

### DIFF
--- a/content/about/faq.md
+++ b/content/about/faq.md
@@ -255,4 +255,4 @@ The tilde {{< keyboard "~" >}} key is a dead key, used to type characters like Ã
 Use the _Keymaps_ preferences. If your preferred layout is not available there yet, you can customize an existing one by drag'n'dropping the keys around.
 
 ### Is there some kind of failsafe mode I can boot into? How do I get to it?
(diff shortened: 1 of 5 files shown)


pulkomandy:
In the code I have used "blocking" or "disabling" instead of "blocklisting". Maybe the documentation should do the same?

humdingerb:
I thought the same, actually. But I read the menu item in the boot options "Blocklist entries" as a verb, what with all other menus there being verbs as well. Maybe we should change that menu as well. How about "Disable components". Components less technical than "entries" and flexible enough to encompass apps, add-ons, perferences, whatever.

korli:
Keep in mind that Beta2 still uses EntryBlacklist entries.

humdingerb:
Right. Damn you, lingering Beta2...
Let's keep EntryBlacklist for now then.

humdingerb:
In the latest commit I'm using "disable" instead of blocklisting. Much better.
If y'alll agree, I'll change the boot options menu to "Disable components" as well.

humdingerb:
If I hear a +1 here, I'll PR the boot menu to say "Disable components" as well...

jt15s:
@humdingerb Yes, I agree with the idea of changing the boot menu too.

humdingerb:
Change to boot options menu label hit the Haiku repo. Merge! Thanks.

PR #489, opened 2021-04-04
Updates coding guidelines

saloniig:
Adds example for empty function

### DIFF
--- a/content/development/coding-guidelines.md
+++ b/content/development/coding-guidelines.md
@@ -79,9 +79,19 @@ Foo::Bar(const char* bar1, T bar2)
 	...
 }
 
-//  Empty functions converted to single line functions
+// Empty functions in class should be converted to inline functions
 
-Foo::~Foo() {}
+class Example
+{
+    void FooFunction() {}
+};
+
+// Empty functions defined outside class should not be converted to inline functions
+
+void
+Foo::FooFunction()
+{
+}
 ```
 
 ```cpp
--- a/content/development/coding-guidelines.md
+++ b/content/development/coding-guidelines.md
@@ -83,7 +83,7 @@ Foo::Bar(const char* bar1, T bar2)
 
 class Example
 {
-    void FooFunction() {}
+		void			FooFunction() {}
 };
 
 // Empty functions defined outside class should not be converted to inline functions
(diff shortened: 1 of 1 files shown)


content/development/coding-guidelines.md
```diff
@@ -78,6 +78,10 @@ Foo::Bar(const char* bar1, T bar2)
 {
 	...
 }
+
+//  Empty functions converted to single line functions
```
pulkomandy, line 82: Should this be allowed outside of the class declaration?

I think this is fine:

```
class Example
{
    void EmptyFunction() {}
};
```

But this isn't:

```
void Example::EmptyFunction() {}
```

In the case it's outside the class declaration it should instead be:

```
void
Example::EmptyFunction()
{
}
```

The case where an empty function is defined outside a class should be quite uncommon anyway. We would usually put it directly in the class declaration.

content/development/coding-guidelines.md
```diff
@@ -79,9 +79,19 @@ Foo::Bar(const char* bar1, T bar2)
 	...
 }
 
-//  Empty functions converted to single line functions
+// Empty functions in class should be converted to inline functions
```
pulkomandy, line 82: "inline" means something specific in C and C++, to avoid confusion we should not use that word here.

```suggestion
// Empty functions in class definitions can be written on a single line
```

content/development/coding-guidelines.md
```diff
@@ -79,9 +79,19 @@ Foo::Bar(const char* bar1, T bar2)
...
 
-Foo::~Foo() {}
+class Example
+{
+    void FooFunction() {}
```
pulkomandy, line 86: Even if it is in a single line, the alignment of the various columns should be preserved.
Also you had spaces instead of tabs for indentation.

```suggestion
		void			FooFunction() {}
```

content/development/coding-guidelines.md
```diff
@@ -79,9 +79,19 @@ Foo::Bar(const char* bar1, T bar2)
...
+};
+
+// Empty functions defined outside class should not be converted to inline functions
```
pulkomandy, line 89:
```suggestion
// Empty functions defined outside class must be formatted the same as other functions
```

pulkomandy:
It looks fine to me now, but I would prefer if another Haiku developer can cross-check this, before we merge it.

PR #509, opened 2021-05-23
Moved Google Code-In Reports to News

panos:
Initial commit, in order to list a couple of articles that I see as worthy of being moved to the News section (Competition results) for long-term posterity reasons, SEO and organization. Many of those sorts of posts from other years were published in the News section and all of this is a bit of a mess that has to be untangled.

I'm only pushing this initial commit before making any further changes in order to make sure that this is fine by everyone.

## TO-DO:

- [x] Add redirects.
- [x] Mildly modify metadata, so that it matches with the metadata of an article. Don't modify more than what is needed.
- [x] Add corresponding notice, see below.

The original content will not be modified, and only a notice reading *This article was originally published on [author]'s blog and moved here for posterity." with a hyperlink linking back to the blog of the author.

### DIFF
--- a/_redirects
+++ b/_redirects
@@ -1,0 +1,8 @@
+# old Google Code-In publications
+
+/blog/scottmc/2010-11-18_2010_google_codein_contest_haiku_selected_participating_organization/	/news/2010-11-18_2010_google_codein_contest_haiku_selected_participating_organization/	302!
+/blog/scottmc/2011-11-10_2011_google_codein_contest_haiku_selected_one_eighteen_participating_organzations/	/news/2011-11-10_2011_google_codein_contest_haiku_selected_one_eighteen_participating_organzations/ 302!
+/blog/scottmc/2014-01-20_google_codein_2013_wrap_report/	/news/2014-01-20_google_codein_2013_wrap_report/	302!
+/blog/scottmc/2015-02-13_google_codein_2014_wrap_report/	/news/2015-02-13_google_codein_2014_wrap_report/	302!
+/blog/digib0y/2017-02-05_gci_2016_wrap_up_report/	/news/2017-02-05_gci_2016_wrap_up_report/	302!
+/blog/digib0y/2018-02-02_gci_2017_wrap_up_report/	/news/2018-02-02_gci_2017_wrap_up_report/	302!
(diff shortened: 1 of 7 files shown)


jt15s (commented):
Might be good to add the message at the top of the reports, and using an alert? You could use the Info alert, just copy paste the following into the file and replace the "Title" and "Body text here" fields: `{{< alert-info "Title" "Body text here">}}`

panos:
> Might be good to add the message at the top of the reports, and using an alert?

This would affect the description. I believe it would be invasive to the original comment as well.

jt15s:
Alright then - we can leave it as it is then.

pulkomandy:
I wouldn't even have included that note. These were meant to be news items from the very start and were just put in the wrong place.

This would be like adding a comment to code saying "this code used to contain a bug, but now it is fixed". That's what the git history is for.

panos:
> I wouldn't even have included that note. These were meant to be news items from the very start and were just put in the wrong place.

Excuse me if I'm making this a bit too difficult, but it just feels a bit dirty on my end to arbitrarily move a blog entry AND remove the credit from the original authors because "I thought it was right". I also did this because of the use of first person in the entries that make no sense whatsoever, as far as the News section is concerned, and that's why I felt like it was a good idea to explain that. Not sure if it should be assumed that everything will work with Git forever either, just because that's how things will work for the foreseeable future.

panos:
I think I should also move the following:
- Icculus Microgrant announcement
- Outreachy announcements

I will take care of that after we deal with the initial changes and sort of agree on how situations like these should be dealt with.

pulkomandy:
Why do these use the first person? If that's intentional and should be preserved, they realy are blog posts and then, why would we be moving them?

(sorry probably I should have re-read them while or before discussing this...)

panos:
> Why do these use the first person? If that's intentional and should be preserved, they realy are blog posts and then, why would we be moving them?

They are announcements that affect the whole organization presented in the form of blog posts from the personal perspective of the writer. Violating the specific style with a note as to why the violation occurs is much better than having this sort of information scattered around, unless if we start proactively archiving events based on correcting tags, which have been abused or used in a non-efficient manner throughout the years.

leavengood:
I think this is fine.

PR #521, opened 2021-06-22
Fixed dark mode issues

jt15s:
Fixed issues in dark mode on the installation guide and the Get Haiku! page.


jt15s:
Would be really appreciated if someone reviewed this!

content/get-haiku/installation-guide.html
```diff
@@ -225,24 +225,24 @@ <h3>Not using GRUB</h3>
...
 <h3>Boot using the CD</h3>
 
-<p>You can also use the boot CD to boot into your Haiku installation without adding it to a boot manager. To do so, hold <span class="key">SHIFT</span> before the CD begins to load. This will bring up the Boot Loader Options.</p>
+<p>You can also use the boot CD to boot into your Haiku installation without adding it to a boot manager. To do so, hold <span class="button">SHIFT</span> before the CD begins to load. This will bring up the Boot Loader Options.</p>
 
 <p><figure role="group">
 	<a id="step-2-1" name="step-2-1"></a>
 	<img src="/files/images/installation_guide/19-bootloader-menu.png" alt="The Boot Loader Menu" />
-<figcaption><p class="step">1</p><p class="caption">To choose the installation to boot from, press <span class="key">ENTER</span> to navigate to the boot volume selection. Your installation should be listed there.</p>
+<figcaption><p class="step">1</p><p class="caption">To choose the installation to boot from, press <span class="button">ENTER</span> to navigate to the boot volume selection. Your installation should be listed there.</p>
```
humdingerb, line 233: Why the change from 'key' to 'button'? Those are all keys...

humdingerb (commented):
Can't comment on the other changes, just spotted the key->button transition and wondered why...

content/get-haiku/installation-guide.html
jt15s: @humdingerb The first part of the documentation used the "button" span, so I just used the "button" span for the whole thing.

jt15s: Actually, never mind - I just realised my mistake - I will try and fix it 😆

jt15s:
@humdingerb I've just restored the key spans.

humdingerb:
OK. I still have no clue about dark mode, so I'm outta here...

content/get-haiku/installation-guide.html
```diff
@@ -225,7 +225,7 @@ <h3>Not using GRUB</h3>
 
 <h3>Boot using the CD</h3>
 
-<p>You can also use the boot CD to boot into your Haiku installation without adding it to a boot manager. To do so, hold <span class="key">SHIFT</span> before the CD begins to load. This will bring up the Boot Loader Options.</p>
+<p>You can also use the boot CD to boot into your Haiku installation without adding it to a boot manager. To do so, hold <span class="button">SHIFT</span> before the CD begins to load. This will bring up the Boot Loader Options.</p>
```
leavengood, line 228: I believe this also needs to remain "key".

leavengood (changes requested):
So how does dark mode work? Is it a browser feature? Can you say what browser you are testing with? Because I would want to make sure this CSS changes don't affect the normal "light mode". Also this needs rebasing after the recent Bootstrap 4 upgrade

pulkomandy:
Yes, the browser has a way to say "I'm using dark mode" and the CSS can query that and use different colors depending on it.
You can test this in recent versions of WebPositive if you have set your Haiku system colors to a dark theme.

In other browsers it's probably an option somewhere in preferences.

jt15s:
Looks like this has a conflict with the shijin4.css file in master - could one of the repo maintainers fix this please and merge if there are no remaining issues? Thanks in advance!

PR #524, opened 2021-06-23
Upgrade Bootstrap to 3.4.1

jt15s:
Upgraded Bootstrap from version 3.3.7 to version 3.4.1 to mitigate a security vulnerability. Since the upgrade is in the same major release, the changes do not impact the website and testing has shown it to function normally.


nielx:
While the security update is probably in the javascript components and not in the CSS, I see no issue with merging this change.

PR #530, opened 2021-07-06
Upgrade to Bootstrap 4 and align some elements

hrnf:
Following the forum topic: https://discuss.haiku-os.org/t/moving-to-bootstrap-5/10905 I spent some time upgrading the website to bootstrap 4. However it's ok if it's decided that currently the upgrade isn't necessary.

In my opinion, the main reason to upgrade is:
Bootstrap 3 was released in 2013. This makes collaboration a bit more difficult because new developers often start working with at least Bootstrap 4.

There are a couple of changes to the UX:
* I aligned some elements to make the website look a bit more clean
* The sidebar in blogposts and pages doesn't go below the content, but above it (the CSS classes which were able to do this are no longer in Bootstrap)
* The style of the navigation bar for mobile devices looks a bit different.
* The way the sizes are calculated (for responsive behaviour) changed a little with Bootstrap 4. In my opinion, it's more pleasant now.


leavengood:
I personally think this is probably a good idea. I would like to see how it looks, which would ideally be using Netlify Deploy Previews, but for certain reasons we have those disabled for branches in forks. But let me see what I can do.

leavengood:
I added your fork as a new remote in my local copy and pushed up your changes into branch bootstrap4. That causes Netlify to create this branch deployment: https://60e52e562d3abb000752ef3d--haiku.netlify.app/

I see the top navigation seems messed up, and also the styling of the sidebars does not look right.

When I run this locally with Hugo I see a different way that the navigation is screwed up, but weirdly the sidebars look fine. I am currently on Windows 10 running Firefox.

Anyhow maybe it needs a bit more work.

leavengood:
One thing to note is for some reasons the URLs for blogs and news item in the branch deploy go to the main website. Maybe it is how Netlify builds the site because locally Hugo does not do that. I can look into it.

hrnf:
> 
> 
> I added your fork as a new remote in my local copy and pushed up your changes into branch bootstrap4. That causes Netlify to create this branch deployment: https://60e52e562d3abb000752ef3d--haiku.netlify.app/

That's weird. I'll look into it. Thanks!

hrnf:
I was able to build the site on Netfly. https://heuristic-mahavira-3d56ef.netlify.app/
It seems the issue was in the config file. There were a couple of variables (baseURL and BuildTypeIsDeploy) that needed to be changed in order to build the site on netfly. I'm pushing a small fix related to the navigation bar anyways.

It's worth noting that I needed to set the Hugo version to 0.83.1. I don't know what version it's used for the site.

themes/shijin4/static/css/shijin4.css
```diff
@@ -604,6 +611,144 @@ footer a, footer span.glyphicon {
 	color: #FF6C00 !important;
 }
 
+/*** BOOTSTRAP 4.0.0 STYLE FIXES ***/
```
waddlesplash, line 614: IMHO this probably belongs in a separate file.

waddlesplash:
Yeah, upgrading to a newer Bootstrap is a good idea. Sorry I don't have time right now to review this in serious detail.

hrnf:
Alright. I'll move the bootstrap 4 fixes into a new file for now when I have time tomorrow.

jt15s:
@hrnf Just looked at your Netlify deploy, and yes, the nav bar is working. The preview looks fine and it seems everything is working okay.

leavengood:
Yeah it looks good now. The changes are subtle but I like it. Thanks for the help!

Once you move the Bootstrap 4 fixes into a new file I'll take a deeper look but most likely we can merge this tomorrow.

hrnf:
Hi all, I pushed the change.
I would like to write a changelog here just for the record. I forgot to add it above. Also, I left the css changes not related to bootstrap in shijin4 (like alignments). I hope that's alright.

Changelog:

* Changes in shijin4.css to align elements
* Added bootstrap4_fixes.css
* Changes in the navbar following bootstrap 4's example
* Changed class input-group-btn to input-group-append as per the documentation example (this doesn't appear in the migration guide)

From the migration guide (https://getbootstrap.com/docs/4.0/migration/):

* Renamed .table-condensed to .table-sm for consistency.
* Renamed .btn-default to .btn-secondary.
* Dropped push and pull modifier classes for the new flexbox-powered order classes. For example, instead of .col-8.push-4 and .col-4.pull-8, you’d use .col-8.order-2 and .col-4.order-1.
* Responsive navbar behaviors are now applied to the .navbar class via the required .navbar-expand-{breakpoint} where you choose where to collapse the navbar. Previously this was a Less variable modification and required recompiling.
* .navbar-default is now .navbar-light, though .navbar-dark remains the same. One of these is required on each navbar. However, these classes no longer set background-colors; instead they essentially only affect color.
* Navbars now require a background declaration of some kind. Choose from our background utilities (.bg-* ) or set your own with the light/inverse classes above for mad customization.
* .navbar-toggle is now .navbar-toggler and has different styles and inner markup (no more three <span>s).
* Dropped the .navbar-form class entirely. It’s no longer necessary; instead, just use .form-inline and apply margin utilities as necessary.
* Explicit classes (.page-item, .page-link) are now required on the descendants of .paginations
* An explicit class, .breadcrumb-item, is now required on the descendants of .breadcrumbs
* Renamed .panel to .card, now built with flexbox.
* .panel-default removed and no replacement.
* Renamed .panel-heading to .card-header
* Renamed .panel-title to .card-title. Depending on the desired look, you may also want to use heading elements or classes (e.g. <h3>, .h3) or bold elements or classes (e.g. <strong>, <b>, .font-weight-bold). Note that .card-title, while similarly named, produces a different look than .panel-title.

themes/shijin4/layouts/blog/single.html
```diff
@@ -1,8 +1,8 @@
 {{ partial "header.html" . }}
 <div id="page">
 {{ partial "navbar.html" . }}
-	<div class="row" id="main-front">
-		<div class="col-md-9 col-md-push-3">
+	<div class="row" id="main-front">
```
leavengood, line 4:
```suggestion
	<div class="row" id="main-front">
```
Trailing whitespace here.

leavengood:
Thanks again for the help!

PR #545, opened 2021-08-04
Replace Google search with DuckDuckGo Search

jt15s:
Replaced Google site search with DuckDuckGo site search for privacy reasons (a lot of Haiku users don't like being redirected to Google).


kallisti5:
Wait! nevermind. I thought this was a blog post or something about it!

I'm ok changing the search bar provider :-D

PR #550, opened 2021-08-17
Gsoc update

saloniig:
GSoC 2021 haiku-format update


scottmc:
the title should include project name to differentiate before the projects

-scott

PR #553, opened 2021-08-19
Fix image not displaying in 20th anniversary news item

jt15s:
Fixed an image URL in the news item as it referenced the wrong file type.


Coldfirex (approved):
looks good to me

PR #557, opened 2021-08-25
Add news item about hiring waddlesplash

leavengood:
Things not really mentioned, which maybe could be added:

- He is getting a competitive rate and will work by the hour.
- Saying "full-time" might not be completely accurate since he plans to work 32 hours per week.
- Current donation levels for private donations need to increase by about 6 to 8 times to make paying him sustainable long term.
- If we get enough donations we could in theory hire more people.
- We have a good amount already saved for paying him (I worry that might reduce donations.)

### DIFF
--- a/content/news/2021-08-25_hiring_waddlespash.md
+++ b/content/news/2021-08-25_hiring_waddlespash.md
@@ -1,0 +1,17 @@
++++
+type = "news"
+title = "Haiku, Inc. has hired an existing contributor to work on Haiku full-time!"
+author = "leavengood"
+date = "2021-08-25 08:15:38-04:00"
+tags = ["haiku", "haiku-inc", "hiring", "full-time", "contractor"]
++++
+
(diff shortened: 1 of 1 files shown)


leavengood:
Here is the Netlify preview: https://612637a97266a00007c3f3c1--haiku.netlify.app/news/2021-08-25_hiring_waddlespash/

waddlesplash:
Saying "approximately" full-time may make sense, then.

Do we need to have my name in here? I know I have it elsewhere (AboutSystem, etc.) but I feel like it's not really necessary in a news announcement; everyone knows me as "waddlesplash".

The rate is not really competitive for the job description. It's competitive enough for the geographic area I'm in, but that doesn't really count.

I would indeed mention that Haiku, Inc. now has a donations target of $80,000 a year (up from $10,000 a year) within the news item itself, and that hiring more people could be possible once that target has been met. I would mention that there is only about a year or two of runway, and that any more depends on a dramatic increase in recurring donations.

leavengood:
I'll make some adjustments but will probably leave the full-time stuff. Changing the wording there is either awkward or belittles how much time you will be putting in compared to most volunteers.

leavengood:
Updated Netlify deploy: https://612642396062d3000739e06d--haiku.netlify.app/news/2021-08-25_hiring_waddlespash/

I swear I set up the integration with GitHub and Netlify but clearly not.

PR #558, opened 2021-08-25
Generate developer/internals documentation

pulkomandy:


jt15s:
Any progress on this? Maybe PulkoMandy can be added to Netlify to make things easier as he can fix things himself.

pulkomandy:
Please, no. I don't want to have to do everything myself. This is not a job for the development team. I don't want to fix this myself.

jt15s:
Hmm, I might have a look then and see if I can fix it.

jt15s:
Okay, looks like it's a Linux-specific build script so I can't really do much on a Windows device.

leavengood:
I have another idea: can we try to find a way to dump this build script? It seems really slow and complicated, and it does a bunch of work, every build, when 99% of the time we are just adding a blog post or something.

Seems like we could tie into GitHub actions or similar in the main repo to rebuild these docs as needed (whenever they actually change, which probably is not that often) and have that push into this repo.

This does not necessarily need to happen in this PR, and I can try to debug the current problem this week, but long term I think it would be good if we can get rid of this build script for another approach. That would also make it so turning on Netlify Previews would be less risky, though I think waddlesplash is overly concerned about the risk on that.

jt15s:
Hmm, maybe we could move the docs into a separate repo? Then the script could run just for that repo, and we can enable Netlify in this repo. Plus it would make the docs easier to edit without having to checkout all the OS source code.

leavengood:
Well the idea, as far as I know, is to have the docs on the website, which is obviously hosted in this repo. My thought is that rather than building the docs on each build when the website is deployed, they are built in some other process and just added to this repo. Then they can be deployed like any other files are, and we can use a normal Netlify Hugo build process.

waddlesplash:
> It seems really slow and complicated, and it does a bunch of work, every build, when 99% of the time we are just adding a blog post or something.

Builds are cached, and the build scripts only rebuild the docs if the Haiku repo actually changed somehow in the interim.

> My thought is that rather than building the docs on each build when the website is deployed, they are built in some other process and just added to this repo. 

I think that would ultimately be more complicated, not less, though I guess it might be simpler for some people. Most of the time we never have to touch the build script though.

leavengood:
> Builds are cached, and the build scripts only rebuild the docs if the Haiku repo actually changed somehow in the interim.

Well the main Haiku repo changes quite a bit (especially with you working) and probably more often than the website changes, yet probably most of the time those changes don't actually affect the various docs being built. So overall I think the current approach has more negatives than positives.

It took me several minutes just now to study the build script to even understand what files it is actually adding. I need to run the script locally to see the total number of files, though I bet it will be much more lightweight to add the resulting files than to clone the entire Haiku repo on every build.

> I think that would ultimately be more complicated, not less, though I guess it might be simpler for some people. Most of the time we never have to touch the build script though.

For me it isn't as much about complication as about:
- improving turn-around time for the website
- making it more clear what the actual documentation files are
- turning on the various nice Netlify integrations without worrying about people trying to hack the build script

leavengood:
Anyhow I'll look into this in the next few days. I'll at least get this script running and debugged to maybe merge this, but also to see the resulting files to gauge how bad it would be to add them to the repo. Then the next step would be to look into how to use GitHub Actions or our build servers to automate keeping things updated.

waddlesplash:
If we are going to add them to the repo, we should at least do it on an alternate branch with a totally separate history and then just simplify the script to merge the branches.

jt15s:
I've always been a bit confused about the fact that documentation is located in the Haiku source itself - would it be possible to move this to another repo and then build it separately to perhaps a subdomain (something like docs.haiku-os.org) or the existing haiku-os.org/docs location if possible? In Netlify I think you can start a new site and just connect the new repo, and that would simplify things by separating the Haiku website (which is powered by Hugo) and the docs (which are powered by Doxygen and Sphinx).

leavengood:
> If we are going to add them to the repo, we should at least do it on an alternate branch with a totally separate history and then just simplify the script to merge the branches.

Why? You think they will be updated so much it will clog the history? I'm not sure. Also maybe you mean another repo? Making a new branch in this same repo with a cleared out history sounds like trouble waiting to happen.

waddlesplash:
The documentation that is in the source itself is largely the API documentation (which of course is closely tied to and partially generated from the source code), and the Userguide. While the former should stay in the tree for obvious reasons, indeed it may make sense to move the Userguide to a separate repository at this point. Sounds like an issue should be filed for that.

waddlesplash:
> You think they will be updated so much it will clog the history

Yes. Or, more to the point, there's simply no reason to clog the main branch with generated files if we can avoid it.

> Making a new branch in this same repo with a cleared out history sounds like trouble waiting to happen.

Nah, it's actually pretty common. The `gh-pages` branch by default in GitHub repositories has a totally parallel history to `master` or any other branch. Git supports it just fine.

leavengood:
Yeah I feel like people have talked about moving the user guide out before. It might even be nice making the user guide its own subdomain like https://userguide.haiku-os.org.

jt15s:
> The documentation that is in the source itself is largely the API documentation (which of course is closely tied to and partially generated from the source code), and the Userguide. While the former should stay in the tree for obvious reasons, indeed it may make sense to move the Userguide to a separate repository at this point. Sounds like an issue should be filed for that.

Ah, that explains why the API documentation is in the source. I'll file an issue on Trac to move the User Guide out to another repo - maybe we could even move it to a subdomain too (maybe userguide.haiku-os.org?)

leavengood:
Great minds think alike 🤣

jt15s:
Ticket filed here: https://dev.haiku-os.org/ticket/17252#ticket

nielx:
> The documentation that is in the source itself is largely the API documentation (which of course is closely tied to and partially generated from the source code), and the Userguide. While the former should stay in the tree for obvious reasons, indeed it may make sense to move the Userguide to a separate repository at this point. Sounds like an issue should be filed for that.

The userguide is maintained here https://i18n.haiku-os.org/userguide/ 
The repository just contains a copy.
It is not unheard of bundling a copy of documentation with the source.

pulkomandy:
I agree the userguide should be removed from the haiku repository. There is another reason to do this: currently the userguide packages are built during the haiku build process and get an hrev number as a version. This means they are regenerated at every change to the haiku sources, and put on the package repository. Which results in user updating haiku downloading the userguide over and over again.

Building it separately (for example at haikuports) and versioning it independently would solve that.

It would also solve the problem that sometimes people will try to submit change requests to fix problems in the userguide, and we have to redirect them to the userguide editor/translator tool instead.

I see no use in having the generated html of the userguide in the Haiku repository. It is not uncommon to bundle the source of the documentation with the sourcecode, but I would say a generated copy of the docs is a lot less common.

So, +1 to moving the userguide out.

The API docs should stay in the Haiku repo and I think the developer docs (what I'm working on here) also should. It makes sense to have this internal documentation easy to edit for developers, and putting it as close as possible to the sourcecode is the best way to achieve that.

waddlesplash:
> Building it separately (for example at haikuports) and versioning it independently would solve that.

I had this on my mind, too, but did not have time to type it out. I agree.

jt15s:
Sidenote: should Haiku Depot Server documentation have its own Doxygen/Sphinx instance or something like that? Currently, it's looking a bit basic, with no search function or anything: https://depot.haiku-os.org/__docs/index.html

PR #560, opened 2021-09-04
Fix href typo in FAQ (Why isn't it called HaikuOS?)

Jorl17:
Hi!

I was browsing through the website and tried to click on the "Why isn't it called HaikuOS" link, realizing it wasn't working. In this commit I fix that typo.

I think that's all there's needed, because I assume none of this is automatically generated.

Thanks


pulkomandy:
Thanks!

PR #563, opened 2021-09-22
Write news item announcing code sprint

jt15s:
Wrote a news item announcing the code sprint, with details about the coding sprint for both existing and new contributors.


pulkomandy:
I have already added a page in the events/conference section. We don't need a news item duplicating the info. Please complete the page in the events section instead:

https://github.com/haiku/website/blob/master/content/conference/2021_online_coding_sprint.md

jt15s:
@pulkomandy I was thinking a news post would be more noticeable since it pops up in the RSS feeds of the site and stays at the top of the site until the next post is created - alternatively I could just move the bulk of the page to the events page and then just have a short page saying refer to that instead? Or would it be better to just do as you said?

pulkomandy:
If the website design doesn't work, let's fix the website design, not hack around it by having the same info in two different places. The website organization is already enough of a mess as it is.

content/conference/2021_online_coding_sprint.md
```diff
@@ -10,19 +10,35 @@ homepage = "https://discuss.haiku-os.org/t/haiku-coding-sprint-2021/11327"
...
+* [Read the development FAQ](https://www.haiku-os.org/development/faq)
+* [Read the Step by Step Guide for new contributors](https://www.haiku-os.org/development/getting-started)
+* [Take a look at our coding guidelines](https://www.haiku-os.org/development/coding-guidelines)
+* [Read the Programming with Haiku series of lessons](https://www.haiku-os.org/development/programming_with_haiku)
+* [Join the development mailing list to stay up to date with the latest formal discussions ](https://www.freelists.org/list/haiku-development)
+* [Have a look around and join the forums](https://discuss.haiku-os.org) - [there's a topic on the coding sprint](https://discuss.haiku-os.org/t/haiku-coding-sprint-2021/11327) where you can post what you plan to work on.
+* Join the Haiku IRC/Matrix channel at `#haiku` (IRC) or `#haiku:matrix.org` (Matrix) where you can talk with other developers and Haiku enthusiasts, ask questions, get feedback and advice. Here, a response to any query is never too far away! Feel free to say hello and don't be afraid to ask if you have any questions.
```
pulkomandy, line 30: If  you want to list bridges, please list all of them (including XMPP and BeShare). There is no reason to do free promotion for Matrix here over the other protocols. Also, you can't just refer to the IRC channel as "#haiku", you need to specify on which server it is, otherwise people won't find it.

Anyway, this will be a lot of useless information here, even after listing all the bridges, people will still have no idea how to connect to these things. So instead, link to the community/irc page, which can go in more length about the details of connecting to the channel. This way, people not familiar with IRC will find all the info they need there.

content/conference/2021_online_coding_sprint.md
```diff
@@ -10,19 +10,35 @@ homepage = "https://discuss.haiku-os.org/t/haiku-coding-sprint-2021/11327"
...
-you want to.
+Anyone can participate, and there is **no obligation on how much time you put in** even though the sprint runs for two weeks. It's a good opportunity for existing contributors to work on something Haiku-related, like fix bugs or embark on that large project that's been sitting on their to-do list for a while, or help resurrect an app (might be their own or an app from the BeOS days sitting around in [HaikuArchives waiting for some TLC](https://github.com/HaikuArchives)), help finalise patches waiting on [Gerrit](https://review.haiku-os.org), submit proposals, etc. etc. And since there'll be more people online during this sprint, you'll have plenty of help from your fellow developers on hand!
+
+## What about new contributors?
+This is also a good chance for new contributors to get started too! Just to ensure you're ready for the sprint, it's a good idea to first check out our [page on how you can get started on development](https://www.haiku-os.org/community/getting-involved/developing), with ideas on what you might like to work on. You might also want to:
+* [Read the development FAQ](https://www.haiku-os.org/development/faq)
```
pulkomandy, line 24: You are drowning the reader with a lot of more-or-less relevant information here. The "getting started" page should already point to all these things, there is no need to repeat it all here.

PR #564, opened 2021-10-19
blog: Some writing on past & current Lenovo products and practices

mmuman:
And a large discussion on right to repair, schematics, datasheets and
specifications, which we all need to write proper drivers.


pulkomandy:
Activity reports are also largely my personal thing. They started as a series of blog posts on WebKit while I was working full time on Haiku, but then widened in scope to cover other people's work. If they should be "official" I'm happy to move them out of my blog.

This is not the website of Haiku inc so there is no problem with that. And I think it is generally in line with the opinion of the Haiku project and the trouble we have getting our software to run on undocumented hardware, and why more people should care about it.

We're not a slick corporate project with controlled communication from the marketing department, we're a group of hackers working on an operating system :)

korli:
This post looks like a continuation/update of the blog post from 2008 (ironically). Anyway what's missing IMO is who the good guys are (even not talking about supporting Haiku).

mmuman:
I suspected there might be some controversy on it belonging here, hence the draft PR. I know there are some "political" statements in there, although it's not far fetched from "we're the good guys, here's why", and just logical thinking. But I understand not everyone might like this. I think there were some discussions on the 2008 one already. Still, we're not only some pet projects, there had been some commercial attempts at using Haiku in the past, and the tied sale discussion sheds some lights on some of the reasons they failed as well, I suppose I could mention this also.

> Perhaps to make it super clear that this is an opinion piece that may not necessarily reflect the opinion of the Haiku Project and Haiku Inc (as a whole) we could put "Opinion:" in front of the post title too?

I'll definitely add a disclaimer in the header at least to make things clear.

> We're not a slick corporate project with controlled communication from the marketing department, we're a group of hackers working on an operating system :)

Yeah and frankly I wouldn't want us to become what I discuss in the post anyway ;-D

> who the good guys are (even not talking about supporting Haiku).

Hmm, you mean like Louis Rossmann, the EFF? I suppose I could take the opportunity to better introduce them in the text. Yeah I didn't formally ask to support Haiku, because well it's done as well as Linux… but maybe…

> Yes, that is true too - maybe I am just being a bit too anxious, being part of the promotion team haha. Maybe this may even get us some press coverage too. And yes, repairability and open-sourcing driver/hardware info is important (though IMO I would have personally toned down the tone in the post, but that's just what I think).

I understand this, again, and welcome the discussion. I don't expect LTT to talk about it to 14M subscribers, but who knows ;-) On the tone, also remember I lived through it, and while I know very well how to write a nice and calm story, this is an experience I had to tell with a minimum of emotions. And it's really a minimum here, I did feel so much relieved seeing this thing go back after a month of bad sleep. Also sometimes when people don't listen to calm voices like those youtubers, shaking them a bit has more effects. It adds to the diversity of ways to tell things.

leavengood:
I just read it myself in about 30 minutes and agree it is fine, and personally it matches a lot of the frustration I feel about modern hardware and I think a lot of people would agree. Especially people interested in Haiku.

korli:
> > who the good guys are (even not talking about supporting Haiku).
> 
> Hmm, you mean like Louis Rossmann, the EFF? I suppose I could take the opportunity to better introduce them in the text. 

Actually I meant which hardware manufacturers do today things correctly, at least some models.

jt15s:
I had a think about this - and you know what? Maybe we can go even bigger with this. Perhaps we (as in the Haiku Project) could adopt the official position that we should have open-source hardware/hardware documentation etc. - the promotion team could do a more official press release plus email it out to our media list. It's good to be part of the movement pushing for open-source hardware/right to repair and whilst we're at it maybe this may help us get noticed too. 

(Plus, we can also consider joining some alliances/movements for open-source hardware and right-to-repair)

jt15s:
@pulkomandy Ah ok, thanks for the clarification!

PR #567, opened 2021-11-28
Add zstd_devel to list of pre-reqs on Haiku

cap10morgan:
Closes https://dev.haiku-os.org/ticket/17431

Looks like most or all of the other host platforms list this as a pre-req

### DIFF
--- a/content/guides/building/pre-reqs.md
+++ b/content/guides/building/pre-reqs.md
@@ -108,7 +108,7 @@ Note that in addition to all the platform-specific packages, you will also need 
 **Basic requirements:**
 
 ```sh
-pkgman install cmd:python3 cmd:xorriso zstd_devel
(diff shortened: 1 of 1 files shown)


content/guides/building/pre-reqs.md
```diff
@@ -108,7 +108,7 @@ Note that in addition to all the platform-specific packages, you will also need
 **Basic requirements:**
 
 ```sh
-pkgman install cmd:python3 cmd:xorriso
+pkgman install cmd:python3 cmd:xorriso zstd_devel
```
korli, line 111: This line is only for Haiku: devel:libzstd would fit better.

cap10morgan: Will do

PR #569, opened 2021-12-07
Improve clarity in compilation guides

panos:
- Add notice that directs people to download the Haiku
  source code, if they have not already done so. This
  will be helpful, as many people open these guides
  directly after looking for them with a search engine.
- Improve headers in 'Compiling for x86_64' guide.
- The PowerPC guide was not updated, as the guide has
  not been updated in a decent amount of time and
  updating it, which would be required in order to
  include the notice I added in the previous documents,
  will require extensive amounts of effort.

### DIFF
--- a/content/guides/building/_index.md
+++ b/content/guides/building/_index.md
@@ -9,6 +9,10 @@ Building Haiku from sources is a surprisingly simple process which allows
 developers to get started quickly developing on Haiku and allowing end users to
 experiment with the latest Haiku sources.
 
+Please ensure that you have obtained a copy of Haiku's source code as described in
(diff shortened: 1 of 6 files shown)


pulkomandy:
Should we move these to the new www.haiku-os.org/docs/develop/ instead? There we could have a table of contents and pages to read in order, possibly making things less confusing?

panos:
Yeah, good idea. Can we hold off from that for now and then do that later, as setting up redirects and fiddling around with links that already have pretty good SEO and have been cross-linked in many different places in bulk may end up badly? That will take some thinking, and what I did should not be the classic "temporary hack that will end up being permanent" to the problem that you want to solve, as I don't believe that what I did here actually solves it.

PR #571, opened 2021-12-07
Update pre-reqs.md for clarity and correctness

madelinecr:
Update bolded "Haiku on Haiku" text to mention checking the article further below for packages required.

Add `bc` to apt dependencies as it is not part of the base Debian 11 install (although it may not be missing on Ubuntu).


pulkomandy:
Do we still need bc? I think it was removed from the build since the introduction of mbrtool for the arm build?

madelinecr:
I cloned the repos today at review.haiku-os.org and followed the x86-64 instructions on Debian 11. Building @nightly-any boot it failed requiring bc. Installing it resulted in a successful build.

I'm not sure about the answer to your question but I'm providing more information in case it's helpful.

pulkomandy:
Ok, thanks, I don't know where it's still used, but we can merge this for now and investigate later. Thanks for updating it!

PR #574, opened 2022-01-10
Add a news item about the financial reports

leavengood:
Not everyone will read the forums and this is probably newsworthy either way.


leavengood:
Here is the Netlify preview, though there isn't much complicated in this post: https://61dba2c0b78ccc0007305db9--haiku.netlify.app/news/2022-01-09_financial_reports/

jt15s (approved):

PR #578, opened 2022-02-04
remove Gumstix and Efika from port status page

davidkaroly:
It seems there's not much activity on the Gumstix or Efika platforms so perhaps we can remove them from the port status?

Unless someone has plans to support them. (even though it might not be so easy especially for the verdex as it seems to have max 128MB RAM according to gumstix.com)


pulkomandy:
That is not the only problem with these, they also use older ARM architectures which require more work (no hard-floats, no atomics on pre-ARMv7 CPUs, etc). So, yes, probably not so great choices of targets for now. Let's remove them and see if someone decides to resume work on them someday :)

davidkaroly:
what about CubieBoard... need-maintainer.png?

pulkomandy:
Possibly. And it would make more sense to target the current one (apparently CubieBoard 6) rather than the old CubieBoard 4?

I will add some new devices there when I can find where I stored them and can test the current state of things on them...

I don't remember who added the CubieBoard to that list, I don't own one.

PR #579, opened 2022-02-04
update status for Cubieboard 4 to need-maintainer.png

davidkaroly:


kallisti5:
This seems fine. I sold my Cubieboard 4 a few years ago.

To be honest, all the individual board maintainers was a lot more relevant when we did per-board builds for arm.
Since we moved to universal builds (relying on rune and other tools to "customize" images for various target devices), this whole section is likely less relevant beyond what individual SoC things we support via fdt. (aka uarts, framebuffers, etc)

PR #582, opened 2022-04-08
replace yaourt with Paru in /content/guides/buidling/pre-reqs.md.

SikkiLadho:
yaourt is dead, use paru instead. 

I have open a ticket on Trac as well. Here's the link: https://dev.haiku-os.org/ticket/17694


kallisti5:
what about yay?  :-)

SikkiLadho:
> what about yay? :-)

Are we gonna argue over this :-) We can always add both!

PR #583, opened 2022-04-08
March Activity & Contract Report: Grammar/Spelling

CodeforEvolution:
It happens to the best of us. :)


CodeforEvolution:
I would take the edits that aren't related to spelling as recommendations. After all, I'm not the original author of the article. :)

pulkomandy:
Looks good to me, but I'll let @waddlesplash and possibly other native english speakers decide about it

waddlesplash:
The spelling fixes are fine, but I think most of the grammar fixes I am not going to take.

I think I learned to put periods inside parenthetical asides, but looking that up I see that's not actually the common convention (maybe it's a bit archaic?)

PR #593, opened 2022-05-30
gsoc2022 introduction blog

MRNIU:
Signed-off-by: Zone.N <Zone.Niuzh@hotmail.com>


MRNIU:
@scottmc @pulkomandy @davidkaroly

pulkomandy:
Ok for me, I'll let the mentors review it in case there's something more to add

davidkaroly:
basically lgtm

one minor thing: perhaps you could be a bit more specific with "run Haiku on ARM" - I assume it's the 32-bit ARM port and main target is running it in qemu? but better write it down explicitly...

MRNIU:
> basically lgtm
> 
> one minor thing: perhaps you could be a bit more specific with "run Haiku on ARM" - I assume it's the 32-bit ARM port and main target is running it in qemu? but better write it down explicitly...

fixed

MRNIU:
@pulkomandy maybe we can merge this?

PR #597, opened 2022-07-02
Up/downgrade guide: Updated instructions

humdingerb:
* Mention SoftwareUpdater as pkgman alternative. The easiest way
  for end-users to stay updated.

* Reversed order of instructions to switch to/from Stable/Nightly.
  Normally, people start switching from Stable to Nightly and only
  later (sometimes) vice-versa.

* HaikuPorts is 'master' for both Stable and Nightly builds.
  Removed the "pkgman add" of the haikuports repo when switching
  Stable/Nightly because they stay the same.

* Use "pkgman full-sync" instead of "pkgman update". Safer when
  switching from Nightly back to Stable.

* Updated repo URL used to downgrade.

Fixes Trac ticket 17818:
https://dev.haiku-os.org/ticket/17818

### DIFF
--- a/content/guides/daily-tasks/updating-system.html
+++ b/content/guides/daily-tasks/updating-system.html
@@ -5,39 +5,40 @@ date = "2014-08-18T13:11:00.000Z"
 tags = ["update"]
 +++
 
-<p>With the introduction of package management you can now upgrade your system in place using the pkgman command.  The update process is straightforward, requires an internet connection, and requires a single reboot. pkgman will handle obtaining the latest updates and applying them to your system.</p>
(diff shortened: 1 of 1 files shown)


waddlesplash:
Wouldn't it make more sense for most of this page to be in the Userguide, with maybe only the specifics about specific-versions remaining as an "advanced task" elsewhere?

humdingerb:
I think we aim to have people use the current official release, and rather release more regularly or 'backport' important changes to the release branch and have people update via SoftwareUpdater.
By adding elaborate instructions to switch to Nightly, we may funnel many 'normal' users to the bleeding edge.

That said, I think a little link to this guide is OK.

How about adding the last part - about permanently downgrading to a specific hrev after a bodged update - to the SoftwareUpdater page (and link from the Boot Loader Options to it).

PR #602, opened 2022-09-06
Update uefi_booting.md

uberjoker:
Small edits to the UEFI Booting Haiku user guides page for clarity.

### DIFF
--- a/content/guides/daily-tasks/_index.md
+++ b/content/guides/daily-tasks/_index.md
@@ -5,7 +5,7 @@ date = "2010-07-10T18:52:35.000Z"
 tags = []
 +++
 
-<p>Haiku is a friendly operating system targeted towards end users and developers alike.  Below is a collection of daily tasks which any new user would need to use Haiku its fullest potential.</p>
+<p>Haiku is a friendly operating system targeted towards end users and developers alike.  Below is a collection of daily tasks which any new user would need to use Haiku to its fullest potential.</p>
 <ul>
 <li><a href='/guides/daily-tasks/install-applications'>Installing applications</a></li>
 <li><a href='/guides/daily-tasks/disable-package-entries'>Disabling components of packages</a></li>
--- a/content/guides/uefi_booting.md
+++ b/content/guides/uefi_booting.md
@@ -19,7 +19,8 @@ Future releases of Haiku should have this process more refined.
 
 ### Partition Layout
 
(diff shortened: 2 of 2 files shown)


pulkomandy:
Thanks :)

korli:
At the moment, the EFI loader always prints "no boot path found, scan for all partitions...", then asks the user to select the boot partition on the SSD. Is there a way to pass the boot partition yet?

pulkomandy:
I don't have this problem on my system, it finds the partition for me. But I don't know which version I'm using exactly, it's a few months older.

There is no way to pass the boot partition because in EFI case, there is no boot partition defined when the bootloader is loaded. The loader is loaded from the EFI system partition without having to select a specific Haiku boot partition.

We could store a partition path in an EFI variable, I think this is how other OS do it. Or maybe we can put some flag in the GPT header of the partition to mark it bootable?

korli:
The GPT partition type was incorrect. After changing to 42465331-3BA3-10F1-802A-4861696B7521 for BFS, the loader found the partition. If more than one BFS partition is present, I suppose it has to be selected manually in the haiku loader.

uberjoker:
> At the moment, the EFI loader always prints "no boot path found, scan for all partitions...", then asks the user to select the boot partition on the SSD. Is there a way to pass the boot partition yet?

> The GPT partition type was incorrect. After changing to 42465331-3BA3-10F1-802A-4861696B7521 for BFS, the loader found the partition. If more than one BFS partition is present, I suppose it has to be selected manually in the haiku loader.

I was seeing exactly the same issue on my install. After reading @korli 's comments, I checked my setup. For some reason, the Haiku partition type was set to Microsoft basic Data, and I had to manually select the partition each time on boot up. I too manually changed the type to Haiku BFS using fdisk and it's all fixed now.

The Haiku installer reformats the partition during install I think - maybe something there is not setting up the type correctly?

pulkomandy:
The partition type is not erased by reformatting, only if you delete and recreate the partition in DriveSetup.

We should add an option in DriveSetup to change the partition type for GPT partition tables, if not already possible, and maybe the Installer should check and warn about it too.

korli:
Looks like a known issue:
see https://dev.haiku-os.org/ticket/17050  https://dev.haiku-os.org/ticket/6287 https://dev.haiku-os.org/ticket/14839

korli:
The gpt partition type can now be changed in Drivesetup.

PR #607, opened 2022-12-19
Add blog article for HTTP Network Services preview in Haiku R1 Beta 4

nielx:
Please review the article. Note that it should not be merged until I did my own proofreading, but I did want to get some more eyes over it.

### DIFF
--- a/content/blog/nielx/2022-12-30_http_network_services_preview_in_r1_beta_4.md
+++ b/content/blog/nielx/2022-12-30_http_network_services_preview_in_r1_beta_4.md
@@ -1,0 +1,111 @@
++++
+type = "blog"
+title = "HTTP Network Services Preview in R1 Beta 4"
+author = "nielx"
+date = "2022-12-30 00:00:01+00:00"
+tags = ["haiku", "software", "development", "netservices"]
++++
+
+The newest beta of Haiku includes a preview of a redesigned, modern HTTP library as the initial part of an renewed Network Services Kit. The primary goal of including the library is to get developers to experiment with it and give feedback on how it works within their software. The secondary goal is to gather some feedback on the use of modern C++ and some additional experimental features. This article provides a background to the new kit, some pointers on how to get started, and some notes on experimental designs that utilize modern C++ features.
+
+<!--more-->
+
+## What's Included in the Preview
+
+The preview that is part of Haiku R1 Beta 4 includes a basic HTTP 1.1 library. It supports the following:
+
+* HTTP 1.1 Requests with:
+    - Standard and custom methods (GET, POST, etc.)
+    - Optional request body with a known size and content type
+    - Custom HTTP Header Fields
+    - HTTP Basic Authentication
+* HTTP 1.1 responses with:
+    - Request headers
+    - Response bodies as a string, or written to specific files
+* Configurable automatic handling of HTTP redirects
+* Configuration of the number of concurrent requests per host and in total
+
+Protocols other than HTTP 1.1, or additional functionality like proxies, cookie handling and other authentication types are not supported yet.
+
+## Getting Started with the Preview
+
+The new HTTP library was first merged in [`hrev56572`](https://git.haiku-os.org/haiku/commit/?id=1111dda699f8093bfdc9b232a919110a1c1b49a7), and is part of the Haiku R1 Beta 4 development package.
+
+There is a specific section in the [API documentation](https://www.haiku-os.org/docs/api/group__netservices.html). The headers can be found in `/boot/system/develop/headers/private/netservices2/` and in order to use the API, you should link against `libnetservices2.a`.
+
+If you use the Makefile engine to build your application, then you should tweak the `LIBS` to add the `libnetservices2.a` library, as well as the applicable C++ runtime libraries (`$(STDCPPLIBS)`) and the network libraries (`libnetwork.so` and `libbnetapi.so`). You also need to add the headers to the `SYSTEM_INCLUDE_PATHS`. For example:
+
+```
+LIBS =  be netservices2 $(STDCPPLIBS) network bnetapi
+
+SYSTEM_INCLUDE_PATHS =  /boot/system/develop/headers/be \
+ /boot/system/develop/headers/cpp \
+ /boot/system/develop/headers/private/netservices2
+```
+
+If you are building your applications on a 32 bit Haiku image, you will have to make sure you use the modern compiler kit. If you are getting unexpected/odd build errors, double check you have run `setarch x86`.
+
+In the nightly images and the master branch, development continues, and therefore there may be incompatible changes to the kit, so be sure to test your software whenever you build it against a newer (unstable) version of the kit. The advice is that if you want to release software based on the preview, then make sure to build it against a known Haiku image. Because the preview functionality is statically linked in your application, it is likely any binaries will continue to work on more modern Haiku builds.
+
+## Experimental Design Features
+
+The Preview of the HTTP Network Services Kit uses some of the C++ language and library features that are currently (mostly) unused in the rest of the project. Most of Haiku follows the design standards originating from Be, which means that most of the API and the code is based on the (pre-) C++98 version of the language. Furthermore, because of Haiku's promise of binary compatiblity, most of the code will still have to build on GCC 2.95.3, which was released on 16 March 2001. This compatibility guarantee is not there for any new APIs and kits. While the new Network Services Kit is in development, it will include some experiments in its design and use of the language.
+
+This section highlights three of those changes: the use of modern C++ language and library features, the introduction of exceptions for error handling, and an experiment in data access and control mechanisms through exclusive borrows.
+
+### Modern C++ (C++17)
+
+| Feature | Example | Description |
+| ------- | ------- | ----------- |
+| **Move Semantics** | [`BHttpSession::Execute()`](https://git.haiku-os.org/haiku/tree/src/kits/network/libnetservices2/HttpSession.cpp?h=r1beta4#n219) | Arguably [move semantics](https://www.cprogramming.com/c++11/rvalue-references-and-move-semantics-in-c++11.html) is the largest new feature in the language in C++11. The libnetservices2 kit utilizes move semantics to the fullest, meaning in the implementation all classes abide by the [rule of six](https://www.modernescpp.com/index.php/c-core-guidelines-constructors-assignments-and-desctructors), which makes it possible to pass objects by value in most cases. The example of the [`Execute()`](https://www.haiku-os.org/docs/api/classBPrivate_1_1Network_1_1BHttpSession.html#ae14a7c52aaddfd69e22be4a8aeff727f) method, is the only place where the new API forces you to pass the request object as an r-value reference. |
+| **Range-based Iteration** | [`BHttpFields`](https://git.haiku-os.org/haiku/tree/src/kits/network/libnetservices2/HttpFields.cpp?h=r1beta4#n477) | The [range-based `for` loop](https://learn.microsoft.com/en-us/cpp/cpp/range-based-for-statement-cpp?view=msvc-170) syntax is supported by the `BHttpFields` class. This allows you to loop through all the header fields in the object using this construct. Internally, it is used in most cases where there are loops through containers. |
+| **String Views** | [`BHttpMethod::BHttpMethod()`](https://git.haiku-os.org/haiku/tree/src/kits/network/libnetservices2/HttpRequest.cpp?h=r1beta4#n70) | The [`std::string_view`](https://en.cppreference.com/w/cpp/string/basic_string_view) object is a very simple concept: it is a pointer to a string, with a length embedded. This makes the use of borrowing (parts of) strings a lot simpler and safer. In the internals of the HTTP parser, it is used to represents certain elements in the HTTP response that are stored in a buffer. This reduced the need for copies. It is worth investigating if it is useful to support string_views in the wider Haiku API, by providing conversions to/from `BString` and as an alternative to `const char*` arguments. |
+| **Optional** | [`BHttpRequest::SetRequestBody()`](https://git.haiku-os.org/haiku/tree/src/kits/network/libnetservices2/HttpRequest.cpp?h=r1beta4#n349) | The [`std::optional`](https://en.cppreference.com/w/cpp/utility/optional) utility allows you to express the situation where something either has a value, or it does not. It is useful both for return values, data structure members, and function arguments. In the example, it is used to handle the situation where sometimes you know the size of the request body, and therefore you can submit an `off_t` size argument, but sometimes you do not, and then you can express it by passing `std::nullopt`. Previously this would have required overloaded methods, or by overloading the meaning of the the size argument so that `0` equates to unknown. |
+| **Smart Pointers** | [`BHttpRequest::SetRequestBody()`](https://git.haiku-os.org/haiku/tree/src/kits/network/libnetservices2/HttpRequest.cpp?h=r1beta4#n349) | C++'s `std::unique_ptr` and `std::shared_ptr` are a great help to reducing data leaks by making it explicit who owns an object on the heap. They are amazing when [used correctly](https://herbsutter.com/2013/05/29/gotw-89-solution-smart-pointers/)! In the example, it is used to move ownership of a data stream (like a file) to the request. If the request then goes out of scope, it automatically cleans up the data stream object. More importantly, smart pointers and their allocation mechanisms (like [`std::make_unique()`](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) ) are used throughout the internal implementation. In fact, you will not be able to find a single `new` or `delete` in the libnetservices2 source code! The Haiku API has [`BReference`](https://www.haiku-os.org/docs/api/classBReference.html) and [`BAutoDeleter`](https://git.haiku-os.org/haiku/tree/headers/private/shared/AutoDeleter.h#n30) utilities, but these do not effectively use move semantics and therefore do not offer the same level of safety as the standard library versions do. |
+| **Scoped Enums** | [`BHttpStatusCode`](https://git.haiku-os.org/haiku/tree/headers/private/netservices2/HttpResult.h?h=r1beta4#n40) | The API uses [`enum class`](https://en.cppreference.com/w/cpp/language/enum) where it makes sense. |
+| **Lambdas** | [`BHttpSession::Request::SendMessage()`](https://git.haiku-os.org/haiku/tree/src/kits/network/libnetservices2/HttpSession.cpp?h=r1beta4#n1052) | [C++ lambda's](https://www.sandordargo.com/blog/2018/12/19/c++-lambda-expressions) allows the definition of anonymous functions. In the example, this is used by a the `SendMessage()` method to allow the caller to add additional information to the `BMessage` that will be sent to an observer/listener. The alternative would be, that the caller of the method would create a BMessage, which would then be enhanced by the `SendMessage()` method. That alternative does pose a dilemma about safe ownership, or copying. This solution circumvents that. |
+
+In addition to these features, the implementation of the kit also uses some other modern language features internally, like [initializer lists](https://en.cppreference.com/w/cpp/language/constructor), the `auto` keyword for variable declaration, `constexpr`, and [string literals](https://en.cppreference.com/w/cpp/language/string_literal).
+
+### Exceptions for Error Handling
+
+The Preview of the HTTP Network Services Kit uses exceptions for error handling, as an alternative to the traditional way of error handling in the APIs inherited from BeOS. In the current public API, error codes are the way to identify and handle errors: functions and methods that may fail have a `status_t` return type, as well as the use of negative values in `ssize_t` or `off_t` return values to signal errors. For many objects, there is an `InitCheck()` method that should be called in order to check if the object is valid and working correctly. For consistency reasons, all new public APIs and most other code do not use exceptions. There is some precedent for experimentation with exceptions, for example in the [private parts of the package kit](https://git.haiku-os.org/haiku/tree/src/kits/package/hpkg?h=r1beta4).
+
+The preview of the HTTP Network Kit requires the user of the API to do all error handling through exceptions. It has the following characteristics:
+* All error handling is done through exceptions. This means that no method will return `status_t` or overload any other return value to signal an error has occurred. It will no longer be necessary to call `InitCheck()` after creating a new object, any issues in the constructor will raise an exception.
+* The API consistently shows which methods and functions can throw exceptions, by the consistent use of the `noexcept` keyword. This is also part of the documentation.
+* The API documentation explicitly states which types of exception can be thrown, as to allow the user of the API to determine which exceptions to handle.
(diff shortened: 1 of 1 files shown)


content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+LIBS =  be netservices2 $(STDCPPLIBS) network bnetapi
+
+SYSTEM_INCLUDE_PATHS =  /boot/system/develop/headers/be \
+ /boot/system/develop/headers/cpp \
+ /boot/system/develop/headers/private/netservices2
+```
+
+If you are building your applications on a 32 bit Haiku image, you will have to make sure you use the modern compiler kit. If you are getting unexpected/odd build errors, double check you have run `setarch x86`.
+
+In the nightly images and the master branch, development continues, and therefore there may be incompatible changes to the kit. This can either mean that your source will no longer build, or the binaries will not work correctly if the implementation of the functionality has changed. If you do decide to release software based on the preview, then make sure to build it against a known Haiku image. Because the preview functionality is statically linked in your application, it is likely any binaries will continue to work on more modern Haiku builds.
```
pulkomandy, line 48: "or the binaries will not work correctly if the implementation of the functionality has changed" seems a bit confusing, I guess it refers to freshly built binaries due to API changes, but it's not very clear and at first glance appears to contradict the next sentence: "it is likely any binaries will continue to work on more modern Haiku builds"

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+| **Range-based Iteration** | [`BHttpFields`](https://git.haiku-os.org/haiku/tree/src/kits/network/libnetservices2/HttpFields.cpp?h=r1beta4#n477) | The [range-based `for` loop](https://learn.microsoft.com/en-us/cpp/cpp/range-based-for-statement-cpp?view=msvc-170) syntax is supported by the `BHttpFields` class. This allows you to loop through all the header fields in the object using this construct. Internally, it is used in most cases where there are loops through containers. |
+| **String Views** | [`BHttpMethod::BHttpMethod()`](https://git.haiku-os.org/haiku/tree/src/kits/network/libnetservices2/HttpRequest.cpp?h=r1beta4#n70) | The [`std::string_view`](https://en.cppreference.com/w/cpp/string/basic_string_view) object is a very simple concept: it is a pointer to a string, with a length embedded. This makes the use of borrowing (parts of) strings a lot simpler and safer, and has a lot of potential in the rest of the API. It is probably a good opportunity to investigate if this can be supported in the Haiku API more (or if it is needed to create a custom variant). |
```
pulkomandy, line 62: BString already has copy-on-write but I don't know if it is agood as what string view does. In any case, we could probably change the internal design of BString, instead of ending up with two classes (string_view and string) in the standard C++, favorizing ease of use of the API over performance

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+| **Smart Pointers** | [`BHttpRequest::SetRequestBody()`](https://git.haiku-os.org/haiku/tree/src/kits/network/libnetservices2/HttpRequest.cpp?h=r1beta4#n349) | C++'s `std::unique_ptr` and `std::shared_ptr` are a great help to reducing data leaks by making it explicit who owns an object on the heap. They are amazing when [used correctly](https://herbsutter.com/2013/05/29/gotw-89-solution-smart-pointers/)! In the example, it is used to move ownership of a data stream (like a file) to the request. If the request then goes out of scope, it automatically cleans up the data stream object. More importantly, smart pointers and their allocation mechanisms (like [`std::make_unique()`](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) ) are used throughout the internal implementation. In fact, you will not be able to find a single `new` or `delete` in the libnetservices2 source code! |
```
pulkomandy, line 64: We already have BReference (similar to shared_ptr) and BAutoDeleter (similar to unique_ptr). These at least deserve a mention and maybe also an explanation of the differences

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+* All error handling is done through exceptions. This means that no method will return `status_t` or overload any other return value to signal an error has occured. It will no longer be necessary to call `InitCheck()` after creating a new object, any issues in the constructor will raise an exception.
+* The API consistently shows which methods and functions can throw exceptions, by the consistent use of the `noexcept` keyword. This is also part of the documentation.
+* The API documentation explicitly states which types of exception can be thrown, as to allow the user of the API to determine which exceptions to handle.
+* The API introduces the [`BPrivate::Network::BError`](https://www.haiku-os.org/docs/api/classBPrivate_1_1Network_1_1BError.html) abstract base class, which defines a standard API for all derived exceptions. Part of this base class is the `origin` property, which holds the location of where the exception occured in order to aid debugging.
+* Next to the exceptions derived from `BError`, users of the API will also find exceptions of the `std::bad_alloc` type for when allocations fail due to the fact that the system is low on memory.
```
pulkomandy, line 79: this seems a bit unpractical to handle. Should we consider using std::nothrow with new and then throwing our own BError derived memory allocation error?

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+
+In the case of the HTTP Network Services kit, the response body can either be stored as the [`BHttpBody::text`](https://www.haiku-os.org/docs/api/structBPrivate_1_1Network_1_1BHttpBody.html#a2c22caaff871835ac25ecafe3789c4d3) string member, or it can be saved to an object that supports the `BDataIO` interface. When the latter option is used, we must account for the fact that that object will then be shared between the thread that created and scheduled the request, and the thread that processes incoming data and writes this to the target. The `BDataIO` interface itself does not presume or implement any locking mechanism, so that means that there should be an additional system to prevent simultaneous data access. In this case we needed a synchronization mechanism with the following properties:
+
+1. The owner should be able to create an object, and manipulate while they have that in their control.
+2. The owner can then exclusively borrow their object to a borrower, who can than manipulate that object. The borrow is exclusive, meaning that the owner can no longer access or write to the object, until the borrower releases it.
```
pulkomandy, line 92: I'm not native english, but I think the verb "borrow" is used in the other direction (the borrower borrows from the owner). And it could be "the owner lend their object to a borrower"

https://dictionary.cambridge.org/grammar/british-grammar/lend-or-borrow

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Thread-safe data access with Exclusive Borrows
+
+One of the biggest challenges in software development is about writing code that handles reading and writing data safely over multiple threads. The main rule is that objects should be owned, read and written by one single thread. As soon as data is shared with more than one thread, there should be mechanisms to synchronize read and write access to the data. The most common methods of synchronization are atomic data structures and semaphores/mutexes. You **need** to employ these synchronization mechanisms, otherwise you will introduce nasty bugs in your code.
```
pulkomandy, line 87: Generally I think this article needs an introduction with a bit of backstory about the previous attempts at libnetservices and their shortcoming.

In particular for the threads and synchronization, I think we tried 3 different ways:
- With user-provided callbacks called from the netservices request threads, letting the user of the API figure out how to do their locking (simpler design, more flexible, but maybe too flexible). People tend to not be so good at locking and maybe it would be better to handle that inside the library
- With fully asynchronous events based on BMessage: easy integration with the existing APIs, but the fully asynchronous system makes it impossible to do some things (such as cancel a request early if the HTTP headers are not what was expected, for example to not display the content of an HTTP redirect). Also BMessage/BLooper proved to have performance problems when using it in this way
- With a mix of asynchronous callbacks where needed, and a generic BDataIO based interface for the data: solved the performance problems from the above, but not the asychronous problem

I'm not sure how the borrowing concept fits into this, my understanding in Rust is that borrowing is related to memory ownership and a way to avoid reference counting and garbage collectors by ensuring ownership at compile time.

In a threading concept, what you describe below is more related to mutexes.

Also, I don't know if that's a good way to go here, I think we are in a relatively simple producer-consumer case, and in that situation, it's possible to do things in many other ways, with atomics and even in almost-lockless ways (no locking needed as long as the buffer is not empty or full).

So, in the end, all we have here is a proper C++ design for a mutex that actually blocks the protected resource from being used, with compile time checks? That's nice in many cases, but not what I would have gone with in this specific one. I don't know if there are future plans here and an attempt to not overdesign things (which is good), if there are, this maybe deserves a mention here.

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
++++
+type = "blog"
+title = "HTTP Network Services Preview in R1 Beta 4"
+author = "nielx"
+date = "2022-12-17 15:31:01+00:00"
+tags = ["haiku", "software", "development", "netservices"]
++++
+
+The newest beta of Haiku will have a preview of a redesigned, modern HTTP library as the initial part of an renewed Network Services Kit. The primary goal of including the library is to get developer to experiment with it, and to give feedback on how it works within their software. The secondary goal is to gather some feedback on the use of modern C++ and some additional experimental features. This article provides a background to the new kit, some pointers on how to get started, and some notes on experimental designs that utilize modern C++ features.
```
humdingerb, line 9:
```suggestion
The newest beta of Haiku will have a preview of a redesigned, modern HTTP library as the initial part of an renewed Network Services Kit. The primary goal of including the library is to get developers to experiment with it and give feedback on how it works within their software. The secondary goal is to gather some feedback on the use of modern C++ and some additional experimental features. This article provides a background to the new kit, some pointers on how to get started, and some notes on experimental designs that utilize modern C++ features.
```

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+
+<!--more-->
+
+## What's included in the Preview
```
humdingerb, line 13:
```suggestion
## What's Included in the Preview
```
If you go for title casing for the section headers...

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Experimental Design Features
+
+The Preview of the HTTP Network Services Kit uses some of the C++ language and library features that are currently (mostly) unused in the rest of the project. Most of Haiku follows the design standards originating from Be, which means that most of the API and the code is based on the (pre-) C++98 version of the language. Furthermore, because of Haiku's promise of binary compatiblity, most of the code will still have to build on GCC 2.95.3, which was released on 16 March 2001. This compatibility guarantee is not there for any new APIs and kits. While the new Network Services Kit is in development, it will include some experiments in its design and use of the language.
```
humdingerb, line 52:
```suggestion
The Preview of the HTTP Network Services Kit uses some of the C++ language and library features that are currently (mostly) unused in the rest of the project. Most of Haiku follows the design standards originating from Be, which means that most of the API and the code is based on the (pre-) C++98 version of the language. Furthermore, because of Haiku's promise of binary compatibility, most of the code will still have to build on GCC 2.95.3, which was released on 16 March 2001. This compatibility guarantee is not there for any new APIs and kits. While the new Network Services Kit is in development, it will include some experiments in its design and use of the language.
```
Maybe even add a newline somewhere, the paragraph is quite long...

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Exceptions for Error Handling
+
+The new HTTP Network Services Kit uses exceptions for error handling, as an alternative to the traditional way of error handling in the APIs inherited from BeOS. In the current public API, error codes are the way to identify and handle errors: functions and methods that may fail have a `status_t` return type, as well as the use of negative values in `ssize_t` or `off_t` return values to signal errors. For many objects, there is an `InitCheck()` method that should be called in order to check if the object is valid and working correctly. For consistency reasons, all new public APIs and most other code do not use exceptions. There is some precedent for experimentation with exceptions, for example in the [private parts of the package kit](https://git.haiku-os.org/haiku/tree/src/kits/package/hpkg?h=r1beta4).
+
+The preview of the HTTP Network Kit requires the user of the API to do all error handling through exceptions. It has the following characteristics:
```
humdingerb, line 74: Above you always capitalize: "The Preview of the HTTP Network Services Kit".

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+* All error handling is done through exceptions. This means that no method will return `status_t` or overload any other return value to signal an error has occured. It will no longer be necessary to call `InitCheck()` after creating a new object, any issues in the constructor will raise an exception.
```
humdingerb, line 75:
```suggestion
* All error handling is done through exceptions. This means that no method will return `status_t` or overload any other return value to signal an error has occurred. It will no longer be necessary to call `InitCheck()` after creating a new object, any issues in the constructor will raise an exception.
```

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+* The API consistently shows which methods and functions can throw exceptions, by the consistent use of the `noexcept` keyword. This is also part of the documentation.
+* The API documentation explicitly states which types of exception can be thrown, as to allow the user of the API to determine which exceptions to handle.
+* The API introduces the [`BPrivate::Network::BError`](https://www.haiku-os.org/docs/api/classBPrivate_1_1Network_1_1BError.html) abstract base class, which defines a standard API for all derived exceptions. Part of this base class is the `origin` property, which holds the location of where the exception occured in order to aid debugging.
```
humdingerb, line 78:
```suggestion
* The API introduces the [`BPrivate::Network::BError`](https://www.haiku-os.org/docs/api/classBPrivate_1_1Network_1_1BError.html) abstract base class, which defines a standard API for all derived exceptions. Part of this base class is the `origin` property, which holds the location of where the exception occurred in order to aid debugging.
```

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+
+The discussion on whether or not exceptions should be used are mostly ideological and fundamental in nature. This preview allows developers to get some hands-on first hand experience of using it in the context of Haiku development, in order to better form their arguments on whether or not the future of the Haiku API is in way of exceptions, or through alternative to `status_t` return values like the [`std::expected`](https://en.cppreference.com/w/cpp/utility/expected) type, which is going to be part of C++23. Or a mixture; there might be good arguments to apply both strategies.
```
humdingerb, line 83:
```suggestion
The discussion on whether or not exceptions should be used are mostly ideological and fundamental in nature. This preview allows developers to get some hands-on first hand experience of using it in the context of Haiku development, in order to better form their arguments on whether or not the future of the Haiku API is in way of exceptions, or through alternatives to `status_t` return values like the [`std::expected`](https://en.cppreference.com/w/cpp/utility/expected) type, which is going to be part of C++23. Or a mixture; there might be good arguments to apply both strategies.
```

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Thread-safe data access with Exclusive Borrows
```
humdingerb, line 85:
```suggestion
### Thread-safe Data Access with Exclusive Borrows
```

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,109 @@
...
+
+In the case of the HTTP Network Services kit, the response body can either be stored as the [`BHttpBody::text`](https://www.haiku-os.org/docs/api/structBPrivate_1_1Network_1_1BHttpBody.html#a2c22caaff871835ac25ecafe3789c4d3) string member, or it can be saved to an object that supports the `BDataIO` interface. When the latter option is used, we must account for the fact that that object will then be shared between the thread that created and scheduled the request, and the thread that processes incoming data and writes this to the target. The `BDataIO` interface itself does not presume or implement any locking mechanism, so that means that there should be an additional system to prevent simultaneous data access. In this case we needed a synchronization mechanism with the following properties:
```
humdingerb, line 89: "In the case of the Preview of the HTTP Network Services Kit,..." ?

humdingerb (commented):
Only a spelling review.
You may want to add a newline here and there for the longest paragraphs.

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
nielx: A `string_view` is a ... view on a string. I.e. it points to the start of the data buffer, and has a length. This mostly makes it a safe(r) and more efficient variant of const char* way of passing pointers to strings around. Additionally, it implements some methods to do some comparisons with std::string objects. I think for reasons of safety and versatility, we should investigate if we want to add the ability to pass a string view like object to methods where now we accept const char*. We should see if we want to reuse std::string_view, or whether there will be a BStringView.

Your point that a BString can be used as an alternative because of the COW design is valid and might be the answer, though one downside is that it still relies on the internal object to be nul-terminated, whereas a string_view by virtue of the fact that it packs a pointer + length does not have that requirement and therefore can also point to substrings within larger data buffers.

Interesting discussion, though for the mailing list/comments on the article. Anything you think I could change in the text to make it clearer?

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
nielx: Will do.

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
nielx: In theory we could, though I think BError needs to be redesigned to make sure it does not allocate anything itself. For example, the `origin` parameter is internally stored as a BString, which admittedly is mostly out of convenience for me, so I think making it const char should be possible. Additionally, the DebugMessage() should be rewritten so that it does not allocate. But it is possible.

In my opinion though, the bigger picture is that there are very few applications that need to handle low/no resource scenarios. In most cases you might as well crash the application in the scenarios. It takes a lot of effort to write resilient code that keeps running during those scenarios. We must do that in the servers and other daemons, as well as Tracker and Deskbar, but beyond that...

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
nielx: Thanks, will review!

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
nielx: To the first point: I will add a bit more details about previous iterations. As for the rest, it feels like I should probably write another article on the particular problem `BExclusiveBorrow<T>` is trying to solve, to work through the alternatives and share why I felt those were not the best solution for this problem.

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
pulkomandy: I am familiar with string_view, thanks.

The problem with std::string is that it has a .c_str() method that returns a const char* pointer to its internal buffer. This means essentially free conversion from std::string to const char*. But it also means the buffer has to be null-terminated at exactly the right place.

You could design a complete API without a thing like std::string and only ever have string_views. If we can, I would prefer that BString goes that way. It means more expensive conversion from BString to const char*, so care must be taken to not need const char* so much. And, like the c++ standard strings, we may be limited by ABI/API stability and decide to preserve both null-terminated and lenght-based strings.

Anyways, future APIs aside, I guess it makes sense to have string_view in an heavily text oriented protocol like http, and it is certainly a better idea than strtok we were using before. So, maybe we can just say that high-performance code here requires specific tools, and BString just won't cut it?

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
nielx: I think that by the time you are running out of resources, even emergency save procedures are perilous and do not have the best odds of succeeding. I think autosave is a much better strategy (and guards against more types of crashes than out of memory). But you're right, another topic for discussion :-)

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
waddlesplash: Yes, "lend" would be correct here, "borrow" sounds wrong.

content/blog/nielx/2022-12-16_http_network_services_preview_in_r1_beta_4.md
nielx: I checked in generated HTML and I think it is manageable. There also is no natural break in this paragraph, so I will leave it as it is for now.

nielx:
Submitting for final review and merge.

content/blog/nielx/2022-12-27_http_network_services_preview_in_r1_beta_4.md
```diff
@@ -0,0 +1,111 @@
++++
+type = "blog"
+title = "HTTP Network Services Preview in R1 Beta 4"
+author = "nielx"
+date = "2022-12-27 00:00:01+00:00"
+tags = ["haiku", "software", "development", "netservices"]
++++
+
+The newest beta of Haiku will have a preview of a redesigned, modern HTTP library as the initial part of an renewed Network Services Kit. The primary goal of including the library is to get developers to experiment with it and give feedback on how it works within their software. The secondary goal is to gather some feedback on the use of modern C++ and some additional experimental features. This article provides a background to the new kit, some pointers on how to get started, and some notes on experimental designs that utilize modern C++ features.
```
pulkomandy, line 9: now that beta4 is out, this "will have" is obsolete

pulkomandy (approved):
Seems OK except for talking about beta4 in future tense in the introduction

content/blog/nielx/2022-12-27_http_network_services_preview_in_r1_beta_4.md
nielx: Updated the intro paragraph (and bumped the publishing date)

PR #612, opened 2023-01-14
get source: improve instructions

n0toose:
- Commands can now be copied and pasted on a variety of systems using user environment variable. As this is not a catch-all solution, a sufficient warning was added.
- Replace example domains with "example.com"
- Improve some headers
- Add some additional background information for better onboarding
- Add additional warning for GitHub pull requests
- Fix English
- Make the tone more informal as to not intimidate new contributors

### DIFF
--- a/content/guides/building/get-source-git.md
+++ b/content/guides/building/get-source-git.md
@@ -101,7 +101,7 @@ git config --global user.email "john.doe@example.com"
 
 Although most contributors tend to disclose their real names, we do not have
 a real name policy and plenty of contributors contribute to Haiku using a pseudonym.
-That means that it is possible for anyone to contribute.
+You are not required to use your "real" name, a pseudonym is fine and several Haiku contributors chose to do that. However, make sure the e-mail address is correct and working, it is what we will use if we need to contact you later.
 
 On macOS, you must set the following option in order to avoid problems with the
 unicode representation of filenames:
--- a/content/guides/building/get-source-git.md
+++ b/content/guides/building/get-source-git.md
@@ -27,7 +27,7 @@ Haiku, you need to follow the instructions for patch submitters in the next
 section.
 </div>
 
-<div class="alert alert-warning">
+<div class="alert alert-info">
 <strong>The buildtools repository is not needed when building from within 32-bit Haiku</strong>.
 In that version, the tools required to build Haiku are included by default.
 </div>
--- a/content/guides/building/get-source-git.md
+++ b/content/guides/building/get-source-git.md
@@ -9,23 +9,27 @@ tags = []
 
 <span class="right"><img src='/images/archive_64.png'></span>
 
-Haiku's source code is currently being hosted in a <a href="http://git-scm.com/" target="_blank">Git based repository</a> with <a href="https://gerritcodereview.com">Gerrit code review</a>.
-Anonymous access will allow anyone to download Haiku's source code; However, the authenticated (non-anonymous) method is required for submitting patches.
+Haiku's source code is currently being hosted on a
+<a href="http://git-scm.com/" target="_blank">Git-based repository</a> using <a href="https://gerritcodereview.com" target="_blank">Gerrit</a>.
 
-Although most contributors tend to disclose their real names, we do not have a real name policy and
-a couple of contributors contribute to Haiku pseudonymously. That means that it's possible for anyone to contribute.
+If you are already familiar with the pull request workflow as used, for example, on GitHub, you
+can find an overview of the differences in <a href="https://gerrit-review.googlesource.com/Documentation/intro-gerrit-walkthrough-github.html" target="_blank">Gerrit user's manual</a>.
 
-<div class="alert alert-warning">
-<strong>The buildtools are not needed when building from within 32-bit Haiku</strong>. Pre-built images of 32-bit Haiku already come with suitable buildtools pre-installed.
-</div>
+Anonymous access will allow anyone to download Haiku's source code; However,
+the authenticated (non-anonymous) method is required for submitting patches.
 
 <a name="anon_access"></a>
 <h3>Git Access - Anonymous testers</h3>
 
 <div class="alert alert-warning">
 <strong>Anonymous access is read-only.</strong> If you want to submit changes to
-Haiku, you need to follow the instructions for patch submitters, in the next
+Haiku, you need to follow the instructions for patch submitters in the next
 section.
+</div>
+
+<div class="alert alert-warning">
+<strong>The buildtools repository is not needed when building from within 32-bit Haiku</strong>.
+In that version, the tools required to build Haiku are included by default.
 </div>
 
 <h4>Build Tools:</h4>
@@ -40,26 +44,67 @@ git clone https://review.haiku-os.org/buildtools
 git clone https://review.haiku-os.org/haiku
 ```
 
-If you don't care about the commit history and want to keep the download small,
-try using the parameter `--depth` when cloning. `--depth 10` limits the history
-to the last 10 commits, for example.
+If you don't care about the commit history and want to limit the download size,
+use the parameter `--depth` when cloning. For example, `--depth 10` limits the
+history to the last 10 commits.
+
+<h4>GitHub</h4>
+
+If your Internet provider limits the websites you are allowed to use, you can use
+our mirror repositories <a href="https://github.com/haiku/haiku" target="_blank">on GitHub</a>
+instead.
+
+They are kept in sync with our main repository, however, these repositories do
+not have any Git tags, which the compiler uses to determine the revision of
+Haiku. To work around this limitation, you can use the
+<a href="https://cgit.haiku-os.org/haiku/tree/build/jam/UserBuildConfig.ReadMe" target="_blank">HAIKU_REVISION build variable</a>
+when building Haiku.
+
+The commands for cloning the repositories from GitHub are:
+
+<h4>Build Tools:</h4>
+
+```sh
+git clone http://github.com/haiku/buildtools.git
+```
+
+<h4>Haiku:</h4>
+
+```sh
+git clone http://github.com/haiku/haiku.git
+```
 
 <a name="dev_access"></a>
 <h3>Git Access - Contributors and patch submitters</h3>
 
+So, you want to contribute to Haiku. That's awesome, thank you for your
+interest! This section will include instructions for setting up your local Git
+environment, an account on our Gerrit instance, as well as some guidelines on
+how to work with other contributors to get your changes included. This guide
+assumes that you are new to contributing to open-source projects. If that
+is not the case, you can skip some of the sections.
+
+<h4>A visual overview of our Git workflow</h4>
+<img src='/files/gitProcess_0.png' alt='A diagram describing the way our Git workflow works. This picture is there for aesthetic purposes and can be skipped. After a repository is cloned locally, if you have an idea, you can convert that idea into a commit using food (a pizza is depicted in the diagram). That commit will receive some change proposals, which will result in some further changes. The circle continues until the change is ready to be included in Haiku itself.'>
+
+<h4>Configuring Git locally<a name="configure_env"></a></h4>
+
 <div class="alert alert-danger">
-<strong>Configure your git!</strong> Before making any commits to the Haiku repository (local even), be <strong>sure</strong> to <a href="#configure_env">configure</a> the Git environment on your local system! Failure to configure git properly before a commit will result in incorrect naming in your commit, making it impossible to give you well-deserved credit for your work.</div>
+<strong>This section is important!</strong> Failing to configure Git properly before a commit will result in an incorrect name being used in your commit, making it impossible to give you well-deserved credit for your work.</div>
 
-<h4>Configure Git on your system:<a name="configure_env"></a></h4>
-
-Before making your first commit on a new system, be <strong>sure</strong> to configure Git. These global settings are stored in your git configuration directory (`~/.git/` or for Haiku: `~config/settings/git/`) and will be appended to <strong>each</strong> commit as your personal information.
+A global Git configuration is stored in Git configuration directory (`~/.git/` or for Haiku: `~/config/settings/git/`) and will be included in <strong>every</strong> commit.
 
 ```sh
 git config --global user.name "John Doe"
-git config --global user.email "john.doe@developers.com"
+git config --global user.email "john.doe@example.com"
 ```
 
-On macOS, you must set the following option in order to avoid problems with the unicode representation of filenames:
+Although most contributors tend to disclose their real names, we do not have
+a real name policy and plenty of contributors contribute to Haiku using a pseudonym.
+That means that it is possible for anyone to contribute.
+
+On macOS, you must set the following option in order to avoid problems with the
+unicode representation of filenames:
 
 ```sh
 git config core.precomposeunicode true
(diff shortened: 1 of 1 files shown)


n0toose:
I spent a lot of time revisiting this again, but my changes may contain some trivial errors or mistakes.

content/guides/building/get-source-git.md
```diff
@@ -48,64 +48,75 @@ to the last 10 commits, for example.
...
+<strong>Configure your Git!</strong> Before making any commits, make <strong>sure</strong> that you have <a href="#configure_env">configured your Git environment</a> properly! Failing to configure Git properly before a commit will result in incorrect naming in your commit, making it impossible to give you well-deserved credit for your work.</div>
 
 <h4>Configure Git on your system:<a name="configure_env"></a></h4>
 
-Before making your first commit on a new system, be <strong>sure</strong> to configure Git. These global settings are stored in your git configuration directory (`~/.git/` or for Haiku: `~config/settings/git/`) and will be appended to <strong>each</strong> commit as your personal information.
+Using `--global` will your personal information in your Git configuration directory (`~/.git/` or for Haiku: `~config/settings/git/`), which will be included in <strong>every</strong> commit.
```
pulkomandy, line 55:
```suggestion
Using `--global` will your personal information in your Git configuration directory (`~/.git/` or for Haiku: `~/config/settings/git/`), which will be included in <strong>every</strong> commit.
```

content/guides/building/get-source-git.md
```diff
@@ -48,64 +48,75 @@ to the last 10 commits, for example.
...
 ```
 
 <h4>Setup an account on Gerrit</h4>
 
-<p>Log in to <a href="https://review.haiku-os.org">Gerrit code review</a>. You currently need a <a href="https://github.com">Github</a> account for logging in.</p>
+<p>Log in to <a href="https://review.haiku-os.org">Gerrit code review</a>. For the time being, using a <a href="https://github.com">GitHub account</a> is necessary.</p>
 
 <div class="alert alert-warning">
-Make sure your ssh key is generated with ed25519. RSA keys will not work with Gerrit.
+Make sure your SSH key is generated using <code class="code">ed25519</code>. RSA keys will <b>not</b> work with Gerrit.
```
pulkomandy, line 73:
```suggestion
Make sure your SSH key is generated using <code class="code">ed25519</code>. RSA keys will <strong>not</strong> work with Gerrit.
```

(for consistency with the other warnings in the page)

content/guides/building/get-source-git.md
```diff
@@ -48,64 +48,75 @@ to the last 10 commits, for example.
...
 </div>
-<p>Upload your SSH public key in <a href="https://review.haiku-os.org/settings/#SSHKeys">Gerrit SSH keys settings</a> page. If you don't have a key yet, you can generate one using ssh-keygen.</p>
+<p>Upload your SSH public key in the <a href="https://review.haiku-os.org/settings/#SSHKeys">Gerrit SSH Keys settings</a> page. If you don't have a key yet, you can generate one using ssh-keygen.</p>
 <p>If the e-mail address used in your commits does not match the one in your github account, you will need to add and verify it in <a href="https://review.haiku-os.org/settings/#EmailAddresses">Gerrit E-Mail address settings</a>.</p>
 
-<h4>Build Tools:</h4>
+<div class="alert alert-warning">
+You may need to replace the <code class="varname">$USER</code> parameter in the following commands.
+</div>
+<p>This can happen if the username on your Gerrit account does not match the username that you use to log in to your computer. This would be the case in Haiku itself, as the default username is <code class="code">user</code>. You can make your life easier by <a href="https://linux.die.net/man/5/ssh_config">modifying your SSH configuration</a>.</p>
```
pulkomandy, line 81: A link to a man page documenting the whole file is not very helpful. Especially in this page which is supposed to be an "all-in-one" getting started tutorial.

content/guides/building/get-source-git.md
```diff
@@ -48,64 +48,75 @@ to the last 10 commits, for example.
...
+<p><code class="varname">$USER</code> is an environment variable that is specific to Unix-like operating systems (e.g. Haiku, Linux, FreeBSD, etc.). If your system does not fall under that category, you will probably have to replace it.</p>
```
pulkomandy, line 82: Well, if that does not work all the time (and especially not with the default Haiku setup), maybe we should not attempt to use it. Let's keep this page as short as possible.

If you want to make things explicit in the page, maybe do something like:

`export GERRIT_USERNAME=john_doe`

(keeping with the user that was used as an example earlier)

content/guides/building/get-source-git.md
```diff
@@ -48,64 +48,75 @@ to the last 10 commits, for example.
...
+<h4>Sending in your first patch</h4>
+
+So, you want to contribute to Haiku. That's awesome, thank you for your interest!
 
-Want to submit your first patch? Thanks, that's great! You don't need to checkout the sources again. Instead you can update your existing copy of the source to use the commiter access. Just change the remote URL:
+<div class="alert alert-warning">
+Pull requests for the Haiku operating system made using GitHub cannot be accepted.
+</div>
+
+If you only have experience with GitHub and downloaded the source code of Haiku from it, don't worry, we've got you.
```
pulkomandy, line 104: I don't think this adds anything relevant to the page and the tone seems a bit too familiar.

There is already a section a bit further down documenting the Gerrit flow for people familiar with Github.

content/guides/building/get-source-git.md
```diff
@@ -48,64 +48,75 @@ to the last 10 commits, for example.
...
 
-The `<login>@` is only needed if your currently logged in username doesn't match your `review.haiku-os.org` username.
-
 ```sh
-git clone "ssh://<login>@git.haiku-os.org/haiku"
-scp -p <login>@git.haiku-os.org:hooks/commit-msg "haiku/.git/hooks/"
+git clone "ssh://$USER@git.haiku-os.org/haiku" && scp -p $USER@git.haiku-os.org:hooks/commit-msg "haiku/.git/hooks/"
 ```
 
-<h4>Switching from read-only to write access</h4>
```
pulkomandy, line 95: this section was clearly labelled and this title made sense if following the previous paragraphs.

Now this info is removed, and there is no context for the commands below.

content/guides/building/get-source-git.md
```diff
@@ -48,64 +48,75 @@ to the last 10 commits, for example.
...
 
-You also should install the <a href="https://review.haiku-os.org/Documentation/user-changeid.html">Gerrit hooks to generate Change-Ids</a>:
+You also should install a Git hook that will automatically <a href="https://review.haiku-os.org/Documentation/user-changeid.html">generate Change-Id's</a> for you:
 
 ```sh
-scp -p <login>@git.haiku-os.org:hooks/commit-msg "haiku/.git/hooks/"
+scp -p $USER@git.haiku-os.org:hooks/commit-msg "haiku/.git/hooks/"
 ```
 
+You should definitely do that, by the way. If a maintainer asks you to correct something later on, Gerrit will use that Change-Id to keep track of your newest changes.
```
pulkomandy, line 118: here as well I think the tone is a bit too familiar. Also, we should really just make this mandatory in Gerrit config, or update our Gerrit to a version that's not broken where the Change-Id is visible on the server side when Gerrit generates it.

content/guides/building/get-source-git.md
```diff
@@ -116,16 +127,16 @@ Haiku's source code needs to reside on a case sensitive file system.
...
 <a name="proxy_access"></a>
-<h4>Getting the source code through an HTTP proxy</h4>
+<h4>Obtaining the source code using an HTTP proxy</h4>
 
 <div class="alert alert-warning">
 Haiku's main Git repository does not allow HTTP access, which is a problem if you are accessing the Internet through a proxy server that only permits HTTP (port 80) traffic.
 </div>
 
-Instead, use one of our mirror repositories at GitHub or Gitorious for anonymous HTTP access, they are both kept in sync with the main repository. First, set Git to connect through your proxy server:
+Use one of our mirror repositories <a href="https://github.com/haiku/haiku">at GitHub</a> for anonymous HTTP access instead. They are both kept in sync with the main repository. First, set Git to connect through your proxy server:
```
pulkomandy, line 136:
```suggestion
Use our mirror repositories <a href="https://github.com/haiku/haiku">at GitHub</a> for anonymous HTTP access instead. They are kept in sync with the main repository. First, set Git to connect through your proxy server:
```

content/guides/building/get-source-git.md
```diff
@@ -137,6 +148,8 @@ git clone http://github.com/haiku/haiku.git
 
 Please note that these repositories do not contain any `hrev` tags, which are used by the Haiku build system to determine the Haiku revision. To work around this limitation, use the <a href="https://cgit.haiku-os.org/haiku/tree/build/jam/UserBuildConfig.ReadMe" target="_blank">HAIKU_REVISION build variable</a> when building Haiku.
 
+The limitation is caused by the fact that GitHub does not support repositories with tens of thousands of tags very well. A new tag is created once every few days and used to create our <a href="https://download.haiku-os.org">Nightly images</a>.
```
pulkomandy, line 151: This is incorrect. Tags are created whenever something is pushed to the master branch, not every few days (it's closer to multiple times a day). And "every few days" wouldn't make sense for nightly images.

Anyway, this kind of technical details should probably not be included in this page. Knowing that it won't work and how to fix/workaround it is enough here. Technical details should go somewhere else?

content/guides/building/get-source-git.md
n0toose, line 82: Should this be removed? The platforms that we support are explicitly stated here: https://github.com/haiku/haiku/blob/07f11ccca916aa828439b25d51713f1cf610b79a/ReadMe.Compiling.md#building-haiku, however, the document does not mention this as it is out of scope.

However, this is the first thing people should supposedly read before they compile Haiku. How do I work around that?

content/guides/building/get-source-git.md
n0toose: This was accidental.

content/guides/building/get-source-git.md
n0toose: > Also, we should really just make this mandatory in Gerrit config.

My changes here were really based after real mistakes I made myself, this is off-topic but you should absolutely go for that.

content/guides/building/get-source-git.md
n0toose: What other mirror repositories are there?

content/guides/building/get-source-git.md
pulkomandy: Really this should be all moved to https://www.haiku-os.org/docs/develop/ where we can have a book-like thing, with pages you have to read in order. The website does not provide that, I think, unless we start manually adding a kind of "next/previous page" link somewhere?

Maybe add yet another warning to this page linking to the list of supported platforms.

content/guides/building/get-source-git.md
n0toose: > This is incorrect. Tags are created whenever something is pushed to the master branch, not every few days (it's closer to multiple times a day). And "every few days" wouldn't make sense for nightly images.

I remembered a conversation we had because of this subject, I'd definitely go "huh, why" and there's no other designated place for that short detail.

content/guides/building/get-source-git.md
n0toose: Understandable. Do I just give the reader a sample ~/.ssh/config and tell them the directories to put them in? (e.g. `~/.ssh/config` on Linux, `/boot/system/settings/ssh/sshd_config` on Haiku.

content/guides/building/get-source-git.md
n0toose: > You should definitely do that, by the way.

is indeed way too direct and friendly, but should the "you" pronoun/addressing the reader be done, in your opinion?

content/guides/building/get-source-git.md
pulkomandy: The answer to "why" is mainly "stupid historical reasons when we switched from SVN".

We could use "git describe --tags" or similar and our life would be a lot simpler. But back when we switched to git, there was no such thing or it was too slow to be usable.

But there could be a page or two of text explaining how the hrevs (and btrevs) work, how they are generated, why they are needed, what problems they create, etc. Then it's not "just a short detail" anymore AND we can provide accurate info :)

n0toose: *misremembered, my bad, I just started contributing again. :D

content/guides/building/get-source-git.md
n0toose: Let's just remove the warning for the time being and hope that the user won't use Windows?

n0toose: P.S. That's a story for another day, but it's a bit hard to move the documents, as they all have to be moved at once and I am not sure what's the standard practice on introducing redirects. (I know where to put them, but has this been done before? It's why I never got around to it.

pulkomandy: we had one confused user try this under WSL2, and it fails only much later. So it would be nice to keep the warning.

n0toose: > we had one confused user try this under WSL2, and it fails only much later.

good god

content/guides/building/get-source-git.md
pulkomandy: If you set the login in the URL as we already recommend, I think the ssh-config is not needed. So it seems simpler to just not mention it at all here?

content/guides/building/get-source-git.md
pulkomandy: I think the simplest way to proceed here is not try to rely on $USER being already set to the right thing.

So let's set it manually:

    set GERRIT_USER=your gerrit username
    git clone $GERRIT_USER@...

And maybe add a screenshot showing where to find the username in Gerrit

content/guides/building/get-source-git.md
pulkomandy: I'm not aware of other mirrors, that's why I just removed the "both" here which referred to github and gitorious before.

n0toose: > Use one of our mirror repositories

... is still there, I'll commit the change and fix this later.

content/guides/building/get-source-git.md
n0toose: Wait, no, "write access" insinuates "commit access", which is why I corrected that header and replaced it with `<h4>Sending in your first patch</h4>`.

n0toose:
Oops, I seem to have been a bit careless with my rebasing as this is a PR. Sorry about that, fixing in a sec.

n0toose:
Fixing few more minor mistakes...

n0toose:
I rebased the changes made in this repository, added `target=_blank` consistently and "butchered" the entire proxy guide and included the anonymous HTTP(S) stuff in the main article (and removed any outdated mentions to just a port 80) so as to make the structure of the document more linear and easy to read. I removed the proxy instructions. I changed the wording of a few additional sentences.

I believe that there's a few factual errors in the piece itself, which I am trying to amend.

n0toose:
Okay, there's still a lot of stuff to fix. Need some more time.

n0toose:
image

This is factually incorrect, isn't it?

pulkomandy:
> image
> 
> This is factually incorrect, isn't it? (because of `review.haiku-os.org`)

It used to be a problem, the HTTP cloning would timeout. I think this has since been solved.

n0toose:
> It used to be a problem, the HTTP cloning would timeout. I think this has since been solved.

Fixed. I just pushed some major changes that changes the entire structure of the document in a way that is linear (and allows for it to be split in two parts later, which was the reason why I claimed that it was not appropriate to include it in `docs/` just yet). I think it looks as good as I want it to be, but I am not too sure about e.g. the way I've placed the overview.

content/guides/building/get-source-git.md
```diff
@@ -9,25 +9,29 @@ tags = []
...
 <h3>Git Access - Anonymous testers</h3>
 
 <div class="alert alert-warning">
 <strong>Anonymous access is read-only.</strong> If you want to submit changes to
-Haiku, you need to follow the instructions for patch submitters, in the next
+Haiku, you need to follow the instructions for patch submitters in the next
 section.
 </div>
 
+<div class="alert alert-warning">
```
pulkomandy, line 30:
```suggestion
<div class="alert alert-info">
```

content/guides/building/get-source-git.md
```diff
@@ -40,141 +44,150 @@ git clone https://review.haiku-os.org/buildtools
...
+<img src='/files/gitProcess_0.png' alt='A diagram describing the way our Git workflow works. This picture is there for aesthetic purposes and can be skipped. After a repository is cloned locally, if you have an idea, you can convert that idea into a commit using food (a pizza is depicted in the diagram). That commit will receive some change proposals, which will result in some further changes. The circle continues until the change is ready to be included in Haiku itself.'>
 
-```sh
-git clone "ssh://<login>@git.haiku-os.org/buildtools" && scp -p <login>@git.haiku-os.org:hooks/commit-msg "buildtools/.git/hooks/"
-```
+<h4>Configuring Git locally<a name="configure_env"></a></h4>
 
-<h4>Haiku:</h4>
+<div class="alert alert-danger">
+<strong>This section is important!</strong> Failing to configure Git properly before a commit will result in an incorrect name being used in your commit, making it impossible to give you well-deserved credit for your work.</div>
```
pulkomandy, line 93: all the sections are... I think having these warnings everywhere just makes the information less concentrated and harder to find.

content/guides/building/get-source-git.md
```diff
@@ -40,141 +44,150 @@ git clone https://review.haiku-os.org/buildtools
...
+git config --global user.name "John Doe"
+git config --global user.email "john.doe@example.com"
 ```
 
-<h4>Switching from read-only to write access</h4>
-
-Want to submit your first patch? Thanks, that's great! You don't need to checkout the sources again. Instead you can update your existing copy of the source to use the commiter access. Just change the remote URL:
+Although most contributors tend to disclose their real names, we do not have
+a real name policy and plenty of contributors contribute to Haiku using a pseudonym.
+That means that it is possible for anyone to contribute.
```
pulkomandy, line 104: While this is an important topic to you, I don't think this is the right place to mention it in this way.

```suggestion
You are not required to use your "real" name, a pseudonym is fine and several Haiku contributors chose to do that. However, make sure the e-mail address is correct and working, it is what we will use if we need to contact you later.
```

content/guides/building/get-source-git.md
```diff
@@ -40,141 +44,150 @@ git clone https://review.haiku-os.org/buildtools
...
 
-<h3>Some Notes</h3>
+<h4>Setup an account on Gerrit</h4>
 
-<h4>Case Sensitive Filesystem</h4>
+<p>Log in to <a href="https://review.haiku-os.org" target="_blank">Gerrit code review</a>. For the time being, using a <a href="https://github.com">GitHub account</a> is necessary.</p>
 
 <div class="alert alert-warning">
-Haiku's source code needs to reside on a case sensitive file system.
+Make sure your SSH key is generated using <code class="code">ed25519</code>. RSA keys will <strong>not</strong> work with Gerrit.
```
pulkomandy, line 118: I would provide a working ssh-keygen example with the right options here. Saying "you can use ssh-keygen but make sure you get an ed25519 key" makes the tutorial not self-contained, and forces people to research ssh-keygen options elsewhere.

content/guides/building/get-source-git.md
n0toose: I honestly just tried to rewrite this, instead of impose my own biases and opinions on the subject. I am committing this for now, but I'll have to change the phrasing later.

PR #620, opened 2023-02-22
Updated the wifi offline download scripts.

wynan:
Noticed that the wifi download scripts did not appear to be working while trying to use them. I've updated them to pull the same files used by the install-wifi-firmwares.sh script. 

I've been able to test the bash script but I don't have a Windows machine at this time to test the bat script.

Should partially address the issues here: https://dev.haiku-os.org/ticket/16612

I was thinking I would update the install-wifi-firmwares.sh script to use b43_fwcutter hpkg after this but let me know if there is a better way and I can make adjustments to the download scripts as well.


static/files/download-data-for-wlan-firmwares.bat
```diff
@@ -20,21 +20,28 @@ if /i {%a4%}=={y} goto :yes
 if /i {%a4%}=={yes} goto :yes
 set HAIKU_TEMP=system\data\firmware
 :yes
-set baseURL=http://cgit.haiku-os.org/haiku/plain/src/system/libroot/posix
+set baseURL=https://raw.githubusercontent.com/haiku/firmware/master/wifi
+set haikuPortsURL=https://eu.hpkg.haiku-os.org/haikuports/master/x86_64/current/packages
```
pulkomandy, line 24: This makes the script usable for only one architecture (here x86_64 was chosen). I think there should be a command line switch to allow getting the x86_gcc2 version instead?

This script is more likely to be useful on older hardware, not sure if many 64bit systems need it.

static/files/download-data-for-wlan-firmwares.bat
```diff
@@ -20,21 +20,28 @@ if /i {%a4%}=={y} goto :yes
...
+
 cd wifi-firmware\%HAIKU_TEMP%\broadcom43xx
-wget http://downloads.openwrt.org/sources/wl_apsta-3.130.20.0
+wget %baseURL%/b43/wl_apsta-3.130.20.0
 cd b43-fwcutter
-wget http://bues.ch/b43/fwcutter/b43-fwcutter-019.tar.bz2
-wget %baseURL%/glibc/string/byteswap.h
-cd bits
-wget %baseURL%/glibc/include/arch/x86/bits/byteswap.h
+wget %haikuPortsURL%/packages/b43_fwcutter-019-2-x86_64.hpkg
```
pulkomandy, line 41: this is specific to x86_64 as well (and same limitation in the .bat file)

waddlesplash: I think the OpenBSD driver for this hardware does not need fwcutter. Perhaps we can switch to that. However, I don't know if anyone actually has the hardware in question anymore to test with?

wynan: Should I just remove this driver then and update the firmware install script? Or does it make sense to remove these scripts entirely? In the end the driver I was hoping would work didn't but I thought I would fix them for the next person.

static/files/alpha4_download-data-for-wlan-firmwares.txt
```diff
@@ -1,56 +0,0 @@
-#!/bin/sh
```
wynan, line 1: This file wasn't referenced on the website as far as I could tell and none of the links seemed to work or were duplicated by the "download-data-for-wan-firmwares.txt".

static/files/download-data-for-wlan-firmwares.bat
waddlesplash: Which driver is that? What is your PCI device? I may be able to switch to an OpenBSD driver which may work better than the FreeBSD ones.

static/files/download-data-for-wlan-firmwares.bat
pulkomandy: Yes, that works too and it's simpler :)

static/files/alpha4_download-data-for-wlan-firmwares.txt
pulkomandy: Thir was for release Alpha 4 (before package management existed), it's ok to delete it now.

static/files/download-data-for-wlan-firmwares.bat
pulkomandy: Should we make the warnings on the website about this script being useful only for extremely obsolete hardware larger? Can we include a cutoff year? Something like "if your wifi card is from after 2006, you don't need this"?

pulkomandy: @waddlesplash can you clarify how OpenBSD avoids fwcutter? If I remember correctly, the issue is that the firmware is built inside the Windows driver and the license does not allow redistribution of modified versions. Did they implement fwcutter inside their driver, or did they just ignore the license and provide the bare firmware file?

waddlesplash: I think they implement something akin to fwcutter inside the driver, yes.

wynan: > Which driver is that? What is your PCI device? I may be able to switch to an OpenBSD driver which may work better than the FreeBSD ones.

Marvell 88W8897 [AVASTAR] 802.11ac - Let me know if you need anything else.

wynan: > Should we make the warnings on the website about this script being useful only for extremely obsolete hardware larger? Can we include a cutoff year? Something like "if your wifi card is from after 2006, you don't need this"?

I like this. Let me try to put some wording together. Might have saved me time.

waddlesplash: > Marvell 88W8897 [AVASTAR] 802.11ac

The drivers this firmware is for all predate 802.11n, much less 802.11ac. There are no FreeBSD or OpenBSD drivers for your hardware, as far as I know.

PR #626, opened 2023-04-22
The Condition Variables Article

waddlesplash:
I have been talking about writing this one on and off for a while now; so, here it is.

DON'T MERGE yet. I put in Monday (04/24) as a prospective publishing date, but let's see how the review goes, first. :)

### DIFF
--- a/content/blog/waddlesplash/2023-04-24_condition_variables.md
+++ b/content/blog/waddlesplash/2023-04-24_condition_variables.md
@@ -1,0 +1,422 @@
++++
+type = "blog"
+title = "Haiku's (Kernel) Condition Variables API: Design & Implementation"
+author = "waddlesplash"
+date = "2023-04-24 13:00:00-04:00"
+tags = ["kernel"]
++++
+
+A few months after my contract with Haiku, Inc. began, I rewrote the implementation of the Haiku kernel's condition variables (as opposed to our userspace condition variables, which are from POSIX.) As this new implementation has run in Haiku for over a year and shipped in the latest release with no sign of any remaining issues, I figured it is high time for a deep-dive on the API, its implementation history, and the design of the new implementation I wrote.
+
+I expect this article will be of broader interest than just to Haiku’s community, because Haiku’s condition variables API has some notable (and powerful) features not found in those of other operating systems, and its implementation is thus likewise unique (at least, as far as I have been able to figure out.)
+
+<!--more-->
+
+## Background
+
+Before delving in to Haiku’s condition variables and how they differ from those of other operating systems, it is first important to understand what they are for, and how they fit in with the other thread-related APIs. (For those already very familiar with atomics and thread synchronization primitives, you can likely skip this section entirely, and jump directly to the section titled “Condition variables”.)
+
+Teams (or in more common terminology, “processes”) are built around *threads*, or more properly, *threads of execution*. Any team will have at least one thread (the “main thread”), and any number of other threads. On systems with multiple CPU cores (and/or “hyperthreads”), multiple threads will be running at once; and even on single-core systems where only one thread will be running at a time, the order in which threads run will not always be the same, since it may depend on disk I/O, system load, etc. Thus, thread operations need to be *synchronized.*
+
+### Atomics
+
+Suppose I have an integer which I want to add 1 to.
+```cpp
+int32* gSharedInteger = new int32(0);
+void add() {
+	int32 localInteger = *gSharedInteger;
+	localInteger = localInteger + 1;
+	*gSharedInteger = localInteger;
+}
+```
+(There are, of course, more compact ways to write the above code in C/C++, but this is closely approximating what the compiler will transform all of them into at build time out of necessity, and thus what will actually run when the code is executed.)
+
+What happens if `add` is invoked for the first time simultaneously in two separate threads? Will `sharedInteger` become `2`, or (if the invocations really occur at exactly the same instant), will it become `1`, since the threads will have “raced” against each other? (There are some helpful tables illustrating this phenomenon [on Wikipedia](https://en.wikipedia.org/wiki/Race_condition#Example).)
+
+To solve this problem, ISAs (instruction set architectures) include _atomic operations_, so called because they act like one operation, not distinct ones. In our `add` function, we performed 3 distinct operations: we fetched the value, added `1` to it, and then set the value. With an atomic operation, these 3 operations become or behave as 1 single operation (for all intents and purposes of the programmer, anyway), and so simultaneous invocations are no longer a problem.
+
+Different compilers, operating systems, etc. expose such ISA-level atomic operations in different ways. On Haiku, we expose them through [a set of functions prefixed with `atomic_`](https://www.haiku-os.org/docs/api/group__support__globals.html#ga8a7c9722b2dcc2d6a92da0d8e6809390) which operate on `int32*`s (and there is also an equivalent set of functions for `int64*`s.) If we use one of them to make our earlier example thread-safe, it will look like this:
+```cpp
+void add() {
+	atomic_add(gSharedInteger, 1);
+}
+```
+(As a bonus, `atomic_add` also returns the value `gSharedInteger` had before `1` was added to it: so it is really a “fetch and add” operation, not just an “add” operation.)
+
+There are a variety of different atomic operations, and many ways to use them. But it is important to keep in mind that these are a ‘low-level’ primitive, and when using them, there are a lot of considerations that must be taken into account: when using multiple atomics in a row, for example, interesting things can happen with operation ordering.
+
+### Spinlocks
+
+Atomic operations mostly serve as a “building block” upon which more ergonomic synchronization primitives are built. Foremost among these, underlying all other thread synchronization primitives in the kernel, is the *spinlock*.
+
+```cpp
+spinlock gLock = B_SPINLOCK_INITIALIZER;
+int32 gSharedInteger = 0;
+
+void add() {
+	acquire_spinlock(&gLock);
+	gSharedInteger = gSharedInteger + 1;
+	release_spinlock(&gLock);
+}
+```
+
+As the name implies, spinlocks simply “spin” (run atomic operations in a loop) until they manage to acquire the lock. They can thus wind up wasting lots of CPU time if another core holds them for a long period of time. (Haiku’s implementation has mechanisms to detect this and trigger kernel panics when it happens.) They are mostly used in implementing other synchronization primitives, and in especially low-level parts of the kernel.
+
+Any thread which is acquiring or holding spinlocks is not allowed to be rescheduled, or even interrupted (by devices, by other CPUs, etc.) for any reason. Thus, spinlocks are always acquired only after the CPU’s “interrupts disabled” mode has been activated, and attempts to acquire them outside this mode will trigger kernel panics.
+
+### Semaphores
+
+The next kind of thread synchronization primitive, and in fact the only high-level one BeOS had in its kernel APIs, is called a *semaphore*, which tracks an abstract “count” and lets threads “acquire” and “release” it. For example:
+
+```cpp
+sem_id gSemaphore = create_sem(1 /* initial count */, "test semaphore");
+int32 gSharedInteger = 0;
+
+void add() {
+	if (acquire_sem(gSemaphore) != B_OK)
+		return;
+	gSharedInteger = gSharedInteger + 1;
+	release_sem(gSemaphore);
+}
+```
+
+Here, I have created a semaphore which starts out having 1 for its “count”. Within `add`, I try to acquire 1 “count” from the semaphore; if there is not one available, the current thread will become *blocked*, and will only be woken up and resume execution once the semaphore can be acquired. Unlike with spinlocks, a blocked thread releases the CPU and lets other threads run while it is waiting; but also unlike with spinlocks, this requires context-switching, which is a relatively expensive operation (though note that “relatively expensive” at this level of OS programming means “takes dozens of *micro*seconds”.)
+
+Note that in the above example, we check the return value of `acquire_sem`, because it can fail: if someone deletes the semaphore while we are trying to acquire it, we will get an error instead of a successful acquisition. (There are also ways to try and acquire semaphores with timeouts, etc. which will open up more possibilities for errors to occur.)
+
+Further note that there is no restriction on how large the “count” can become in a semaphore. While in this example, the `acquire` and `release` calls are strictly balanced, I could have, e.g., a producer-consumer scenario, where every time the ‘producer’ thread created some resource, it `release`’d the semaphore, and likewise the ‘consumer’ thread would wait on resources to be created by calling `acquire`.
+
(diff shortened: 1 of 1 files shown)


waddlesplash:
@axeld, @weinhold: I know you designed the original API here, perhaps you might have some time to look through this article and see if you have any comments?

@Midar: I know you're not very active on Haiku at the moment, but I think you know a thing or two about condvars on other OSes, so you may have some idea of how "novel" Haiku's API is, and also my new implementation of it. (Is there any chance this is "original research", and not just a neat algorithm?)

@zdykstra: I know we talked about condvars in the past, and you seem like a good candidate for a Haiku user who's not a kernel developer, and is somewhat removed from kernel development (though I guess @humdingerb is, also) to review this article while it's still in draft form, and see what parts you can follow, which can be improved, and which are completely "out there" :)

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+
+Before delving in to Haiku’s condition variables and how they differ from those of other operating systems, it is first important to understand what they are for, and how they fit in with the other thread-related APIs. (For those already very familiar with atomics and thread synchronization primitives, you can likely skip this section entirely, and jump directly to the section titled “Condition variables”.)
+
+Teams (or in more common terminology, “processes”) are built around *threads*, or more properly, *threads of execution*. Any team will have at least one thread (the “main thread”), and any number of other threads. On systems with multiple CPU cores (and/or “hyperthreads”), multiple threads will be running at once; and even on single-core systems, the order in which threads run will not always be the same, since it may depend on disk I/O, system load, etc. Thus, thread operations need to be *synchronized.*
+
+### Atomics
+
+Consider if I have a variable which I want to add 1 to.
+```cpp
+int32 gSharedInteger = /* initializer... */;
```
waddlesplash, line 25: should be ` = 0`

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+}
+```
+What happens if `add` is invoked for the first time simultaneously in two separate threads? Will `sharedInteger` become `2`, or (if the invocations really occur at exactly the same instant), will it become `1`, since the threads will have “raced” against each other?
+
+To solve this problem, ISAs (instruction set architectures) include _atomic operations_, so called because they act like one operation, not distinct ones. In our `add` function, we performed 3 distinct operations: we fetched the value, added `1` to it, and then set the value. With an atomic operation, these 3 operations become 1 single operation (for all intents and purposes of the programmer, anyway), and so what order threads run them in is irrelevant.
+
+Different compilers, operating systems, etc. expose such ISA-level atomic operations in different ways. On Haiku, we expose them through a set of functions beginning with `atomic_` which operate on `int32*`s (and there is also an equivalent set of functions for `int64*`s.) If we use one of them to make our earlier example thread-safe, it will look like this:
+```cpp
+void add() {
+	atomic_add(gSharedInteger, 1);
```
waddlesplash, line 39: should be `&gSharedInteger`

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+
+This is very different from the API found on other operating systems, or in POSIX, in which one “waits” directly on the condition variable itself, and so there must be a single lock to switch from. On Haiku, there does not need to be a lock at all; since (for example) the operation resulting in the condition variable being notified can be started *after* the `Entry` is `Add`ed to the `Variable`, making it impossible for any race to even theoretically occur.
+
+(For example, in Haiku’s kernel, creating a new team (or `fork`ing an existing one) creates a `Variable` and an `Entry` waiting on it, before it even begins creating or forking the team, and then only invokes `Wait` right at the end of the process.)
```
waddlesplash, line 139: disambiguate: "of the procedure".

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+```
+(The precise details of how `mutex_switch_lock` manages to acquire the second lock and release the first lock without having any periods of time where either both locks are held, or neither lock is held, shall have to be a story for another time.)
+
+Having discussed all these details, now we can discuss condition variables.
+
+## Condition variables
+
+What is a ‘condition variable’? Put in the terms we have already discussed, it is a “countless semaphore.” In other words, a condition variable is simply an object that has no state of its own besides knowing what threads are ‘waiting’ on it, and can be told to ‘notify’ one or more of those threads with a value, waking them up.
+
+Besides not having an explicit “value”, it also does not need a global ID (besides its address in memory). Thus it is much cheaper to create and destroy than semaphores are, and there is also no hard limit on how many there can be at any point in time. (To be fair, some kinds of semaphores on other operating systems may not need global IDs either, but there are still other differences I shall not deal with in this article.)
```
waddlesplash, line 114: singular-plural mismatches

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+Finally, we have `fWaitStatus`. This value is either positive (if the `Entry` has been `Add`ed somewhere, but not yet notified and woken up), or `<= 0` (once it has been.) When `Wait` is invoked, it will either return this value or a timeout-related error code, depending on what happens. It has some relevance for synchronization, but not very much, as we will (once again) see later on.
+
+#### `Add`ing `Entry` to a `Variable`
+
+This is the most straightforward action of all: the `Variable`’s `fLock` is acquired, and then the following actions are taken inside the `Entry`:
+```
+inline void
+ConditionVariableEntry::_AddToLockedVariable(ConditionVariable* variable)
+{
+	ASSERT(fVariable == NULL);
```
waddlesplash, line 190: all other assertions later on omitted for brevity's sake, this one probably can be too.

Midar:
My experience with condition variables on various operating systems is mostly in abstracting them into a single API that works everywhere. It turns out that you can have basically the POSIX API everywhere. Anything that goes further usually is not workable on all OSes. For reference, this is the API I came up with that I made work so far on POSIX, Windows and AmigaOS annd as you can see, it's basically an object-oriented variant of the POSIX API: https://github.com/ObjFW/ObjFW/blob/master/src/OFCondition.h

The API I see here seems to be more of an equivalent of Windows's event API and `WaitForMultipleObjects` (which on POSIX-like systems you could do via `eventfd` and `select`/`poll`/`kevent`/`epoll`). Windows's events are actually a way to implement POSIX-style conditions on Windows, so they are more generic: https://github.com/ObjFW/ObjFW/blob/master/src/platform/Windows/OFPlainCondition.m

So the big question is: Do you want something that just works on Haiku, or something that could also be made to work on other OSes? If it's just for Haiku, software will probably not use it. If you want such an API to find more widespread use, I would recommend to do it differently: Add `eventfd` (Linux and the BSDs already have it) that returns a file descriptor that can the be stuffed into `select`/`poll`/`kevent`/`epoll`. There can then be a C++ library on top of it with your API, which would make it much more likely for software to pick this up. This library could also use Windows events on Windows.

But in any case, I would recommend to make `eventfd` available first for existing POSIX software before adding something new.

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+
+Consider if I have a variable which I want to add 1 to.
+```cpp
+int32 gSharedInteger = /* initializer... */;
+void add() {
+	int32 localInteger = gSharedInteger;
+	localInteger = localInteger + 1;
+	gSharedInteger = localInteger;
+}
+```
```
pulkomandy, line 31: maybe explain that this is not how a developer would typically write this code, but it is what the compiled code will do at a low level, and that's why a race exists

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+What happens if `add` is invoked for the first time simultaneously in two separate threads? Will `sharedInteger` become `2`, or (if the invocations really occur at exactly the same instant), will it become `1`, since the threads will have “raced” against each other?
+
```
pulkomandy, line 33: I would include a sequence diagram showing how this can go wrong, but maybe this is not the point of this article? Can we link to something that has additional details, a wikipedia page or so?

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+Different compilers, operating systems, etc. expose such ISA-level atomic operations in different ways. On Haiku, we expose them through a set of functions beginning with `atomic_` which operate on `int32*`s (and there is also an equivalent set of functions for `int64*`s.) If we use one of them to make our earlier example thread-safe, it will look like this:
+```cpp
+void add() {
+	atomic_add(gSharedInteger, 1);
+}
+```
+(As a bonus, `atomic_add` also returns the value `gSharedInteger` had before `1` was added to it: so it is really a “fetch and add” operation, not just an “add” operation.)
+
+There are a variety of different atomic operations, and many ways to use them. But it is important to keep in mind that these are a ‘low-level’ primitive, and when using them, there are a lot of considerations that must be taken into account: when using multiple atomic operations in a row, interesting things can happen with memory ordering, for example.
```
pulkomandy, line 44: a link to the documentation for atomic operations in the haiku book (or if there is none, to the header file where we define them) would be nice here

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+void add() {
+	acquire_spinlock(&gLock);
+	gSharedInteger = gSharedInteger + 1;
+	release_spinlock(&gLock);
+}
+```
+
+As the name implies, spinlocks simply “spin” (run atomic operations in a loop) until they manage to acquire the lock. They can thus wind up wasting lots of CPU time if another core holds them for a long period of time. (Haiku’s implementation has mechanisms to detect this and trigger kernel panics when it happens.) They are mostly used in implementing other synchronization primitives, and in especially low-level parts of the kernel.
+
+Any thread which is acquiring or holding spinlocks is not allowed to be rescheduled, or even interrupted for any reason. Thus, spinlocks are always acquired only after the CPU’s “interrupts disabled” mode has been activated, and attempts to acquire them outside this mode will fault.
```
pulkomandy, line 63: I think understanding this requires a specific definition if "interrupted". I understand that it's about hardware interrupts, but it would be better to make this explicit?

Likewise, "fault" is not great here since it can mean a variety of things. In this case, talking about a kernel panic would avoid confusion (for example with CPU level faults like segmentation faults).

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+
+### Semaphores
+
+The next kind of thread synchronization primitive, and in fact the only high-level one BeOS had in its kernel APIs, is called a *semaphore*, which tracks abstract “units” and lets threads wait for them to become available. For example:
```
pulkomandy, line 67: I think we can call them "tokens" instead of "units"? That doesn't change the meaning, but I think it makes things just a little less abstract and easier to reason about

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+
+void add() {
+	if (acquire_sem(gSemaphore) != B_OK)
+		return;
+	gSharedInteger = gSharedInteger + 1;
+	release_sem(gSemaphore);
+}
+```
+
+Here, I have created a semaphore which starts out having 1 “unit”. Within `add`, I try to acquire a “unit” from the semaphore; if there is not one available, the thread in question will become *blocked*, and will only be woken up and resume execution once the semaphore can be acquired.
```
pulkomandy, line 81: Maybe make it explicit that there is an upside and downside compared to spinlocks:

Upside: a blocked thread will release the CPU for other threads to run while it is blocked
Downside: this means there will be a context switch when the thread is blocked, and that is a relatively slow operation

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+
+If an `Entry` is destroyed without having been woken up by the `Variable` it was added to, it must remove itself from that `Variable` before it becomes invalid. Thus, in `~ConditionVariableEntry()`, if `fVariable != NULL`, a method is invoked to perform the removal (which is also used in another condition, as we will see later.)
+
+It is possible that, while the `Entry` is trying to remove itself from the `Variable`, that the `Variable` is also trying to notify and remove this very same `Entry`. This section is thus considered “critical”, and so we begin by entering interrupts-disabled mode, even before we try to acquire the `Variable`’s spinlock.
+
+```cpp
+void
+ConditionVariableEntry::_RemoveFromVariable()
+{
+	InterruptsLocker _;
```
pulkomandy, line 211: this syntax deserves some explanation for people not used to C++ code and our strange naming convention for these variables?

content/blog/waddlesplash/2023-04-24_condition_variables.md
```diff
@@ -0,0 +1,416 @@
...
+```
+
+The next thing to be done is fetch the `fVariable` pointer and store it in a local variable. We do this with an atomic operation most just as a precaution, as a plain load or store of the CPU’s native pointer size (or any smaller value) is likely to be inherently “atomic”, anyway.
```
pulkomandy, line 214: while this is fine on modern x86, other architectures may really need the atomic, for example due to having multiple CPU sockets without automatic cache coherency (as in the BeBox) or weaker memory orderings (needing specific "barrier" instructions). In these cases it's not really "just a precaution"

humdingerb (commented):
I'm sorry, that's just too low level for my little code-dabbling mind. After reading about half of it, my eyes just glazed over... :)
FWIW, there weren't any egregious grammatical or spelling errors that I could detect in the first half, so from that perspective it looks good. :)

content/blog/waddlesplash/2023-04-24_condition_variables.md
waddlesplash: Wikipedia has a diagram of sorts, indeed. I'll just link to it.

content/blog/waddlesplash/2023-04-24_condition_variables.md
waddlesplash: We don't appear to have our own documentation for this. The Be Book calls it the "thread count", which seems confusing. I'll just refer to it as the "count".

content/blog/waddlesplash/2023-04-24_condition_variables.md
waddlesplash: Right. Deleted this phrase, as explaining that is likely beyond the scope of this article.

waddlesplash:
All comments addressed.

> Do you want something that just works on Haiku, or something that could also be made to work on other OSes? 

This is just the kernel-level API used for Haiku-native device drivers. We already have the POSIX API in userland, and an equivalent of it in kernel-land, for that matter. I added comments to the article clarifying this.

PR #630, opened 2023-05-11
Add post showing icon sets proposed in icon set contest

Zardshard:


humdingerb:
Theres a stray file static/files/blog/zardshard/icon_sets/out-0 there that's the same as 575.png.

Zardshard:
Good catch!

pulkomandy:
I did some mailing list archive digging.

The contest was announced here: https://www.freelists.org/post/haiku/Icon-Artwork,11

The discussion and submissions continue for a few months (06/2006 to 11/2006) but there is a convenient summary thread once a winer has been picked: https://www.freelists.org/post/haiku/Icons-summery-of-past-discussions

Maybe these links can be added to the blogpost so we don't have to do the digging again

Zardshard:
Nice! I also included some changes I wanted to make.

PR #632, opened 2023-05-24
Post "How to find memory leaks"

Zardshard:


content/blog/zardshard/2023-05-23_how_to_find_memory_leaks.md
```diff
@@ -0,0 +1,59 @@
++++
+type = "blog"
+title = "How to find memory leaks"
+author = "Zardshard"
+date = "2023-05-23 10:44:38-04:00"
+tags = ["malloc_debug", "libroot_debug.so", "heap", "memory", "debugging"]
++++
+
+In 2010-2011, mmlr created a new memory allocator: the guarded heap memory allocater. This allocator help detect various bugs such as writing past the end of allocated memory, reading uninitialized memory, and freeing freed memory. These uses are detailed in ["Using malloc_debug to Find Memory Related Bugs"](https://www.haiku-os.org/blog/mmlr/2010-02-08_using_malloc_debug_find_memory_related_bugs/). Later, in 2015, mmlr had a new project: updating the memory allocator to be able to report memory leaks.
```
diversys, line 9: memory allocater (typo)

content/blog/zardshard/2023-05-23_how_to_find_memory_leaks.md
```diff
@@ -0,0 +1,59 @@
...
+        <_APP_> IconEditorApp::IconEditorApp() + 0x1f6
+        <_APP_> main + 0x27
+        <_APP_> _start + 0x3f
+        0xf821926ae5 (lookup failed: Invalid Argument)
+        0x7ffefb24d258 (lookup failed: Invalid Argument)
+```
+
+These, then, are two leaks accompanied by a stack trace giving directions to where in the code the memory was allocated. Unfortunately, these do not have line numbers associated with them. For example, `BMenuItem::SetLabel(char const*) + 0x4c` has a `+ 0x4c` instead of a line number. This is the offset to the assembly instruction immediately _after_ the assembly instruction in question.
+
+The Debugger can be used to find what line of source code `BMenuItem::SetLabel(char const*) + 0x4c` is referring to. First, get the disassembly of the function in question. Second, add the offset (`0x4c` in this example) to the address of the first instruction in the function. Third, set a breakpoint to the instruction _before_ the instruction indicated by the offset. Finally, switch back to source code view. The line highlighted by the breakpoint is the line that called the next function in the stack trace or allocated the memory.
```
pulkomandy, line 59: I think you can do this using the `addr2line` command instead. I'm surprised that leak_analyzer doesn't already do that?

Zardshard: It doesn't appear the `addr2line` command works. I tried it with a simple "Hello World!" program on Haiku and Linux. Linux gave a file and line number while Haiku gave `??:0`.

Zardshard: `llvm-addr2line` works, however, it cannot accept input along the lines of `main+0x3`. I have to first find the base address of the function, e.g. `0x986`, then add `0x3` to it. Let's see if there is any better way.

Zardshard: No better way that I could find. I'll just leave the article as written.

content/blog/zardshard/2023-05-23_how_to_find_memory_leaks.md
```diff
@@ -0,0 +1,59 @@
++++
+type = "blog"
+title = "How to find memory leaks"
+author = "Zardshard"
+date = "2023-05-23 10:44:38-04:00"
+tags = ["malloc_debug", "libroot_debug.so", "heap", "memory", "debugging"]
++++
+
+In 2010-2011, mmlr created a new memory allocator: the guarded heap memory allocator. This allocator help detect various bugs such as writing past the end of allocated memory, reading uninitialized memory, and freeing freed memory. These uses are detailed in ["Using malloc_debug to Find Memory Related Bugs"](https://www.haiku-os.org/blog/mmlr/2010-02-08_using_malloc_debug_find_memory_related_bugs/). Later, in 2015, mmlr had a new project: updating the memory allocator to be able to report memory leaks.
```
diversys, line 9: This allocator help detect (helps)

Zardshard: Good catch! I ran it through a grammar checker and found two more errors.

content/blog/zardshard/2023-05-23_how_to_find_memory_leaks.md
waddlesplash: addr2line only works if the file in question has DWARF information, of course. `libbe` does not by default.

Zardshard: Would `libbe` not having DWARF information affect `addr2line` on a "Hello World" program that only uses the C++ standard libraries?

waddlesplash: The C++ standard libraries also do not have DWARF information, I don't think. Anyway you can likely use `awk` plus `objdump` to turn symbol names into addresses and add them together (or just use a Python script) and then use `llvm-addr2line` on top of that.

Zardshard: Indeed, that is the method I had come up with, except I used `nm` and `bc`. Now, I can either (a) create a script to automate this process, (b) put these instructions into the article, or (c) leave the article as written. I had decided to go with option c.

waddlesplash: I think it makes sense to write a short script and embed it in the article.

Zardshard: Good idea!

pulkomandy: Really this should be fixed in libroot_debug or leak_analyzer, they should be able to show line numbers if Debugger can do it?

PR #634, opened 2023-05-30
Update activity.js

bunnyDasari:
Added arrow function in place of regular function to increase readability over regular function. and added strict equality(===) === compares if the values and the types are the same in the place of ==

### DIFF
--- a/static/js/activity.js
+++ b/static/js/activity.js
@@ -14,7 +14,7 @@ const DISPLAY_ITEMS_COUNT = 7;
 
 var tabs = document.querySelectorAll("#activity-tabs li");
 for (var i = 0; i < tabs.length; i++) {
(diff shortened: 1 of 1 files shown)


waddlesplash:
These changes seem mostly cosmetic, I'm not sure there's a necessity for them...

Zardshard:
My (limited) experience is that the style being proposed tends to be preferred. Using strict equality (===) instead of normal equality can help prevent bugs. Arrow functions are slightly shorter than writing out `function`. In the end, though, it's all mostly cosmetic and doesn't really matter either way.

waddlesplash:
Looking at the account which opened this PR, it seems they opened others like this one across other repos. Between that and the fact that these changes are mostly cosmetic, let's just close.

PR #637, opened 2023-06-10
Self-host Noto Sans and Open Sans

nipos:
Let's finally fix something that has bothered me since I first discovered Haiku years ago...
There's no need to feed data-kraken Google with more data everytime someone visits the Haiku website.
Also,it's considered a violation of the European GDPR law: https://blog.runcloud.io/google-fonts-gdpr/
While I know that Haiku has it's legal home in the US,it's still nice for an open-source project to comply with privacy law ;)
Furthermore,reducing unneeded third-parties on the website strengthens the independence of the project,which I think is one of Haikus goals.
As website hosting (and traffic) is sponsored by Netlify,I don't see any downside with self-hosting this.


pulkomandy:
Do we really need this many font files?

- What is Open Sans used for?
- Are all these font formats really useful or can we remove some of them?
- Do we need both the '400' and '700' Noto variants?

(the change is otherwise welcome, but let's see if we can cut some weight from it while we're at it?)

nipos:
Open Sans is used for the #preface and .title according to shijin.css
Noto Sans is needed in both 400 and 700,because 400 is the normal weight for standard text and 700 is bold,for example used in ul.nav-tabs
The different font formats are required to support all browsers.
Most support woff2 or at least woff nowadays,but as the site has pretty good backwards-compatibility otherwise,I wouldn't break that by not supplying supported font formats.
The browser only downloads one of the formats when it loads the page,the one that it supports best.

nielx:
Are there any attribution requirements from the licenses of the fonts?

nipos:
Both fonts are licensed under SIL Open Font License 1.1,which you can read here: https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL
In their FAQ they state that attribution is not required: https://scripts.sil.org/cms/scripts/page.php?item_id=OFL-FAQ_web#e71fabc0

nielx:
I will do a test build to validate it but it looks good to me so far.

nielx (approved):
Looks good to me. Loads the fonts on my machine (Edge; latest version), though could use a bit more widespread testing.

PR #656, opened 2023-09-24
Add additional examples in the code style guide

andponlin:


content/development/coding-guidelines.md
```diff
@@ -348,8 +348,48 @@ const uint32 kOpenFile = 'open';
...
+
+#include <Message.h>
+
+
+class ThisClass {
+// ...
+};
+
+
+#endif	// JWT_TOKEN_HELPER_TEST_H
```
waddlesplash, line 374: incorrect guard

content/development/coding-guidelines.md
```diff
@@ -348,8 +348,48 @@ const uint32 kOpenFile = 'open';
...
-*   In header files, there is no blank line between the license text and the header guard.
-*   After the copyright header (including the header guard in header files), there are exactly two blank lines before the rest contents.
+*   In header files, there is no blank line between the license text and the header guard but following that should be two blank lines. There should also be two blank lines before the close of the header guard as shown below;
+
+```cpp
+/*
+ * Copyright 2004-2007 Haiku, Inc. All rights reserved.
+ * Distributed under the terms of the MIT License.
+ *
+ * Authors:
```
waddlesplash, line 358: This block is optional, really. I think in newer files it's started to fall by the wayside. But I guess we can leave it in these guidelines...

content/development/coding-guidelines.md
```diff
@@ -348,8 +348,48 @@ const uint32 kOpenFile = 'open';
 ```
 
 *   In some cases you might have to extend the copyright list of an existing file. Instead of using the alternative scheme like in the latter example the **preferred** method is to add a copyright line for "Haiku, Inc." and then list authors like in the first example.
-*   In header files, there is no blank line between the license text and the header guard.
-*   After the copyright header (including the header guard in header files), there are exactly two blank lines before the rest contents.
+*   In header files, there is no blank line between the license text and the header guard but following that should be two blank lines. There should also be two blank lines before the close of the header guard as shown below;
+
+```cpp
+/*
+ * Copyright 2004-2007 Haiku, Inc. All rights reserved.
```
waddlesplash, line 355: `,` after year

content/development/coding-guidelines.md
```diff
@@ -348,8 +348,48 @@ const uint32 kOpenFile = 'open';
...
+ * Distributed under the terms of the MIT License.
+ *
+ * Authors:
+ *		Jonathan Smith, optional@email
+ *		Developer Name, optional@email
+ */
```
waddlesplash, line 387: Do we need to repeat the copyright block in all the examples?

content/development/coding-guidelines.md
andponlin: Thanks; fixed.

content/development/coding-guidelines.md
andponlin: I think that it is helpful because often times I want to see how many spaces are to be left between the copyright block and the start of the guard or header includes.

content/development/coding-guidelines.md
andponlin: Done

waddlesplash (approved):

PR #659, opened 2023-11-20
style: improve color contrast for link

blacksponge:
This PR attempts to improve links accessibility.

Changes
* Improve color contrast
* Add underlines for links

### DIFF
--- a/themes/shijin4/static/css/shijin4.css
+++ b/themes/shijin4/static/css/shijin4.css
@@ -26,6 +26,7 @@ body {
 
 a {
 	color: #336698;
+	text-decoration: underline;
 }
 
 #page {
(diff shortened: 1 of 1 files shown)


waddlesplash:
Considering how many inline links we use, I don't know how desirable this change is.

Won't users who want links to always have underlines use various options in their web browser, anyway?

pulkomandy:
Links should always be underlined, that's the default behavior and there is no reason to change it.

I would like @nephele-gh to review the color changes and make sure both the light and dark modes are still OK, before I merge this.

pulkomandy:
I'm merging this since it is an improvement over what we have now, further changes can be made in later commits if needed :)

Thanks for your contribution!

blacksponge:
Thanks, and sorry, I haven't had much time to come back to this since I made the PR. I will consider @nephele-gh suggestions for a follow-up PR but I will first have to check if the colors pass the 4.5:1 recommended contrast ratio in both light and dark mode.

waddlesplash:
I think the addition of underlines to all links really does make the site look worse.

Bootstrap removes them by default, and plenty of websites stick with that default with no problem. Can't we do the same, please?

waddlesplash:
> The link colors in safari for me are now too blue, the previous light blue was better visible.

Agreed.

> Having them on the sidebars in which every entry is a link is debateable though.

Yes, at least there they can definitely be removed.

humdingerb:
I never had a problem with links not being underlined, but just being significantly differently styled than normal text, i.e. bolded and coloured. As long as they get underlined on mouse-hover, to really make it clear this is a link when the user is about to click on it. A nice feedback.

PR #661, opened 2024-02-05
Coding Guidelines: C-style casts

nielx:
Recently proposed to codify the C-style cast style as having no space after the operator. This will be codified in `haiku-format` and in our written guidelines as well.

See: https://discuss.haiku-os.org/t/haiku-style-format-rfc-c-style-casts-and-spaces/14557/1


pulkomandy:
You added an example, but you didn't mention in the text that there should be no space. After all the effort to clarify this, I think it would be a good idea to write it down explicitly?

nielx:
The document itself does not spell out every rule and uses examples to express them in places, but I don't mind adding it explicitly!

nielx:
Updated: checked that the cast operator is indeed an operator (at least according to the C-specification), so added it at that part of the document.

waddlesplash (approved):

PR #662, opened 2024-02-06
Blog Post: RFC Class Member Declaration Formatting

nielx:
This change also uses the <blockquote> tag, so introduce CSS support for blockquotes. I am not a CSS wizard, so feel free to suggest improvements.

Light mode:

image

Dark mode:

image

### DIFF
--- a/content/blog/nielx/2024-02-05_rfc_coding_guidelines_formatting_class_member_declarations.md
+++ b/content/blog/nielx/2024-02-05_rfc_coding_guidelines_formatting_class_member_declarations.md
@@ -1,0 +1,286 @@
++++
+type = "blog"
+title = "RFC Coding Guidelines: Formatting Class Member Declarations"
+author = "nielx"
+date = "2024-02-05 16:13:58+00:00"
+tags = ["haiku", "rfc", "development", "coding-guidelines"]
++++
+
+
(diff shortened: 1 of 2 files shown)


content/blog/nielx/2024-02-05_rfc_coding_guidelines_formatting_class_member_declarations.md
```diff
@@ -0,0 +1,286 @@
...
+
+### Why the Current Rule Introduces Inconsistencies
+
+The current rule introduces a table-like format for class members. In the (quick) historical search, I did not find the exact origins of this style. From my perspective, it works particularly well under the following conditions:
+
+- It is used in simple, non-nested classes and structs with (mostly) member functions and member data.
+
+- It works best when the class merely declares its members, but does not define them. This works best when there is a clear separation between a definition of a class with the declaration of its members in the header file, and a definition of the members in the source file.
+
+- It works well with standard Haiku's C++98 style, which does not use namespaces, nested classes or exceptions. There is also very limited use of templated types.
```
pulkomandy, line 84: these things are not really related to C++98

namespaces are used (mainly the BPrivate one, but there is some more in the package kit). I don't remember what our rules are for namespaces currently, but for example in WebKit, declarations in namespaces are not indented, so this isn't a problem.

Namespaces, nested classes, templates and exceptions are all available in C++98.

(My main point here is to remove the C++98 mention, we can discuss the other aspects of this probably in the comment of this article once it is merged).

nielx: Thank you for the proofread. The point of that particular construct was to clarify that the formatting works well if you stick to a particular subset of the language, but I see how it can be construed as an incorrect point (and I will be hammered for it), as well as that the point is made better in other parts as well, so I will take it out.

waddlesplash:
Is the blog really the place for such RFCs, though?

nielx:
> Is the blog really the place for such RFCs, though? (I would think the forums or development mailing list is the right place.)

I don't know if there is a better place for it. It seems a bit too long for a discourse post or mailing list post.

pulkomandy:
> Is the blog really the place for such RFCs, though?

Developers are free to use their blogs however they want. It is the least official/formal part of the website. So it seems a good place to start a discussion about something.

PR #663, opened 2024-02-06
Fix url for git commit hook

nephele-gh:


korli:
There are at least two other occurences of https://git.haiku-os.org/tools/hooks/commit-msg

PR #671, opened 2024-03-16
Compiling: Improve U-Boot instructions for RISCV64

n0toose:
I have been experimenting with RISC-V and decided to try it with Haiku. I faced a few difficulties when doing so, and decided to enrich the documentation with a few resources for other people in my position (as explaining how to compile U-Boot is probably beyond the scope of Haiku, but pointing people to the right direction isn't).

### DIFF
--- a/content/guides/building/compiling-riscv64.md
+++ b/content/guides/building/compiling-riscv64.md
@@ -19,9 +19,9 @@ Building the RISCV64 compiler toolchain is quite easy using Haiku's ```configure
 
 > For a complete list of flags for the configure script, see [Haiku's Configure Options](/guides/building/configure)
 
-1. Perform a git clone haiku and buildtools
+1. Obtain a copy of [Haiku](https://cgit.haiku-os.org/haiku/) and its [buildtools](https://cgit.haiku-os.org/buildtools/).
 2. Within the haiku source directory, create your workspace for RISCV64 via ```mkdir generated.riscv64; cd generated.riscv64```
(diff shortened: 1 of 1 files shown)


kallisti5:
We actually offer pre-compiled u-boot binaries for a wide selection of devices here:  

https://github.com/haiku/firmware/tree/master/u-boot

Here's the qemu command I use to test Haiku (and build haikuports for riscv64:

```
qemu-system-riscv64 -m 6G -M virt -parallel none -kernel ~/Code/firmware/u-boot/riscv64/qemu/u-boot.bin -netdev user,id=testnet,net=192.168.76.0/24,dhcpstart=192.168.76.9,hostfwd=tcp::5555-:22 -chardev vc,id=serial,logfile=/tmp/test-riscv64-serial.mon,signal=off -serial chardev:serial -device ati-vga -global virtio-mmio.force-legacy=false -device qemu-xhci -device usb-tablet -device usb-kbd -device virtio-blk-device,drive=x0,bus=virtio-mmio-bus.0 -device usb-net,netdev=testnet -drive file=~/haiku-riscv64.qcow2,format=qcow2,if=none,id=x0 -smp 2
```

n0toose:
The documentation is unfortunately missing lots of context that has iteratively built up over the years (and because it is canonical, it arbitrarily and greatly limits the amount of people that can actually contribute here - myself included), and I would find it very hard to catch up by myself in one shot.

I cannot guarantee that I will completely overhaul its content, but the plan in my head was to add a few notes about EDK II before doing anything else. The documentation should also optionally include instructions about Internet access and HaikuPorts (`virt-manager` could help here). Is the current state of this PR any better?

PR #673, opened 2024-04-02
legacy-docs/bebook: cancel_timer() takes a timer* parameter, not a timer_t*

dalmemail:
See https://github.com/haiku/haiku/blob/master/headers/os/drivers/KernelExport.h#L178


pulkomandy:
Unfortunately, we are not allowed to update the be book due to licensing problems (it is available under a "no derivative" creative commons license only).

That's why we are rewriting the documentation at https://api.haiku-os.org
You can submit documentation for this there if it isn't available already.
The source for this documentation is found in Haiku main repository at https://cgit.haiku-os.org/haiku/tree/docs/user

PR #678, opened 2024-05-11
NVMM GSoC 2024 Initial Post

dalmemail:


dalmemail:
BTW, is there anyway to tell it that my forum username is dalmegnu and not dalme? Without changing the blog name

waddlesplash:
> BTW, is there anyway to tell it that my forum username is dalmegnu and not dalme? Without changing the blog name

I don't think there is, no...

PR #685, opened 2024-05-28
Post GSOC status update

Zardshard:


pulkomandy:
Thanks for the writeup about this, it clarifies a few things, and repeats some things that you had told me but I already forgotten.

I think I should clarify why I'm trying to avoid OpenGL if possible.

It's not necessarily a performance concern. In fact, for some operations, it may very well be faster than software rendering.

The problem is, it's quite a lot of work to enable it because:

- It uses an abstraction library (ANGLE) that does not know about Haiku
- There is no knowledge of Haiku BGLView in the code. Some parts look hardwired to use EGL which is a more standardized way to do what BGLView does: "connect" OpenGL with actual windows and suffaces that can be shown on screen. Adding that may require some work
- In addition to the complexity of interactions between WebProcess and MiniBrowser, there are also interaction with app_server to take into account. We need to implement this with bitmaps that are shared, not only between the two applications, but also with app_server who will be the one doing both the drawing (in one application) and the displaying (in another). I think I can understand how this would work with just BBitmap: app_server is the only one actually accessing the bitmaps, and the two apps are just passing them around, and sending drawing commands. But if OpenGL is involved, we add a new layer of problems, because OpenGL is running on the client side (not in app_server). This opens the possibility of many more synchronization problems.

That being said, if it turns out to be the simplest way, maybe we can go with it, see where it gets us, and work on the other parts of getting WebKit2 running. And if it turns out to be slow, or complicated, or have some other problem, we can always try to disable it later.

In the case of WebKitGTK (and, I think, all other WebKit ports), all the drawing is done client-side in WebProcess. There is no server-side drawing as app_server does. So we are in a bit of a different situation here.

I don't know if you want to integrate that info in your blog post, otherwise, I'll just repost it as a comment in the forum after the post is published.

Zardshard:
I'm not going to integrate it, but that is part of the reason why I posted the blog post: so that you can know where I currently am.

Zardshard:
Can you post it as a comment on my blog post? I think it will be helpful to clarify future progress (I'm already dealing with EGL)

PR #691, opened 2024-06-29
Post "Fixing IPC in WebKit"

Zardshard:


Zardshard:
Waddlesplash suggested using BMessage::ReturnAddress, or something similar. I want to investigate that.

pulkomandy:
Hello,

A few clarifications:

- BMessenger is already somewhat bidirectional: it's possible to send a reply to a received BMessage. But this has a limitation, that a reply can only be sent after you have received a message.
- In Rajagopalan's initial implementation in 2019, the UI Process had to keep a map from "identifiers" to the corresponding messengers. This would allow to identify a connection with just the key into that map (an integer), and then later fill in the details (creating BMessenber instances, etc). If possible, I would like to avoid that. It should be possible to work with BMessengers directly, and pass them around. In the current version of the branch, I have changed that so that the BMessenger itself is serialized into the command line sent to each of the children.
- One difficulty fitting BMessenger into the existing WebKit architecture is that you can only create a BMessenger for an already existing BHandler, and only from within the process/team that runs that handler (with one exception, you can target the main BHandler of another application). UNIX socketpairs don't have this restriction, so you can have the UI process create a socketpair, pass it on to one of its children, and then immediately start sending messages to it, for example. In our case, it has to go in the other direction: UI process creates a BMessenger to itself, pass that to the child process, and the child process then has to send a first message with a "here is how to reach me" information.
- Even more complex in the case of the Network Process to Web Process communication: I think in this case, the UI process creates a socketpair and then sends each end of it to the two children. And then the children start talking with each other.

None of this is really a problem in absolute terms, once we have a communication between UI process and its children, we can build things up from there. But, we may have to do things in a significantly different order compared to the UNIX version, for example, which may require some changes to cross-platform code that starts the processes (or not using the cross-platform classes and using our own).

(again after writing this I realize maybe some of it will be better as a forum reply to your article once it is posted)

Zardshard:
> \- Even more complex in the case of the Network Process to Web Process communication: I think in this case, the UI process creates a socketpair and then sends each end of it to the two children. And then the children start talking with each other.

Actually, I made [an image](https://github.com/haiku/website/pull/691/files#diff-4bc7d5c8cf2275c28692e40e248a8d21f994980b3916345d4b017263264b8211) showing how this process works normally. It shouldn't be too hard to use the existing process.

> (again after writing this I realize maybe some of it will be better as a forum reply to your article once it is posted)

lol, true :smile:

Zardshard:
Is this ready to be merged or are you still working on reviewing it?

pulkomandy:
I didn't get a notification that it wasn't draft anymore. Merged now :)

PR #697, opened 2024-07-28
blog/trungnt2910: Debugging progress 2

trungnt2910:
Second report, recording a lot of significant implementation details.

### DIFF
--- a/content/blog/trungnt2910/2024-06-28_[gsoc_2024]_debugging_progress_2.md
+++ b/content/blog/trungnt2910/2024-06-28_[gsoc_2024]_debugging_progress_2.md
@@ -1,0 +1,483 @@
++++
+type = "blog"
+title = "[GSoC 2024] Improving the Userland Debugging Experience - Progress Report #2"
+author = "trungnt2910"
+date = "2024-06-28"
+tags = ["haiku", "software", "gsoc", "gsoc2024", "gdb", "gdbserver"]
++++
+
+# Project status overview
+
+## Completed tasks
+
+The GDB port is feature-complete. A recipe has been submitted to HaikuPorts.
+
+This took a bit longer than expected due to complexities in building the full GDB compared to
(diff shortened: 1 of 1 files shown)


content/blog/trungnt2910/2024-06-28_[gsoc_2024]_debugging_progress_2.md
```diff
@@ -0,0 +1,485 @@
...
+On `ptrace`-based OSes, this is done by handling the `SIGCHLD` sent to the tracer every time a
+tracee hits an event. The signal handler writes a pipe, marking the FD as ready for the next `poll`
+in the event loop.
+
+Haiku does not send the debugger `SIGCHLD` or similar signals. Furthermore, ports are used instead
+of FDs for communicating with the debugger. A potential solution is to put all open debugger ports
+into an `event_queue`, then return the file descriptor of that queue to the event loop. However,
+this file descriptor type on Haiku currently
+[does not](https://github.com/haiku/haiku/blob/2d24dd21846e3d71961c5d84190d235397c0d9ae/src/system/kernel/events/event_queue.cpp#L599-L600)
+support polling (please upvote [this ticket](https://dev.haiku-os.org/ticket/18954) if you are
```
Comment (waddlesplash) [inline content/blog/trungnt2910/2024-06-28_[gsoc_2024]_debugging_progress_2.md:433]: This is a pretty technical detail and turned out to not be necessary to port GDB's non-stop mode. I don't think it makes sense to ask users to upvote the ticket. Linking it and mentioning the discussion may make sense, at most.

Comment (trungnt2910) [reply content/blog/trungnt2910/2024-06-28_[gsoc_2024]_debugging_progress_2.md]: Removed the call to upvote that ticket, but:

> turned out to not be necessary

Still necessary to cleanly port it. The current solution is more like a temporary hack.

Spawning threads in otherwise single-threaded programs have implications, especially in a program that uses `fork` here and there. Normally I would avoid threads as much as possible and use an event-based asynchronous approach (either signals or polling).

> This is a pretty technical detail

The whole blog is meant to be a lot of technical details...

PR #699, opened 2024-08-10
blog/diegoroux: playback showcase.

diegoroux:
Video showcasing that playback **is** working.


korli:
Showing the Media preferences with the playback device would have been nice.

PR #711, opened 2024-08-26
Post "Debugging when the Debugger doesn't work right"

Zardshard:


content/blog/zardshard/2024-06-20_debugging_when_the_debugger_doesnt_work_right.md
```diff
@@ -0,0 +1,76 @@
++++
+type = "blog"
+title = "Debugging when the Debugger doesn't work right"
+author = "Zardshard"
+date = "2024-08-26 8:54:30-04:00"
+tags = ["tutorials", "debugger"]
++++
+
+Well, while working on HaikuWebKit, I've had to wrangle with the Debugger a lot. As you may know, Debugger, is, well, buggy. This blog post documents the workarounds I've developed for two bugs: Debugger failing to resolve a function name, and Debugger's disassembly of a function being incorrect.
```
pulkomandy, line 9: That way of putting it is not very nice to the developer of Debugger. Especially this:

> As you may know, Debugger, is, well, buggy.

WebKit is certainly testing the limits of debuggers, and the use of lld may be part of the problem as well.

Also, are there bug reports about these two problems? You could create them if they don't exist, and link to them in the post. This will make things a bit more constructive and there is some hope that someone will look into it :)

Zardshard: > That way of putting it is not very nice to the developer of Debugger. Especially this:
>
>>    As you may know, Debugger, is, well, buggy.

Heh, yeah, now that you mention it...

> Also, are there bug reports about these two problems? You could create them if they don't exist, and link to them in the post. This will make things a bit more constructive and there is some hope that someone will look into it :)

Hmm, I managed to make a minimal example that reproduces both of those problems. Yes, everything goes back to using lld. If these problems only occur when using lld, then, I wonder if my blog post is very useful in the first place. Well, let me go file the bug(s)...

Zardshard:
Alright, I filed https://dev.haiku-os.org/ticket/19021 and https://dev.haiku-os.org/ticket/19022. Since the bugs only happen when using lld, I think the probability of users stumbling on this problem is pretty low, so I don't think it's worth having its own blog post.

PR #718, opened 2024-09-19
Use correct mount point in UEFI Install Guide.

infrastation:


diversys:
I think we can drop quotes here to make commands less noisy.

PR #720, opened 2024-12-27
pre-reqs: Add Void Linux to the supported build hosts

Tevo45:
I always end up having to re-figure out what I need to install in order to build Haiku, figured it might be worthwhile to add it to the guide. Tested on glibc x86-64, don't know about musl, or other architectures. ARM packages are based on the requirements for other distributions, but I didn't try building it.


waddlesplash:
Thanks!

PR #724, opened 2025-06-20
gsoc/ideas: add an idea for ufshci

korli:


pulkomandy:
We should create the 2026 page first, and preserve the 2025 one as it was, since GSoC 2025 is already in progress (but we were not selected)

pulkomandy:
Added the paragraph to the 2026 page. Thanks!

PR #763, opened 2026-09-21
Revise Flash support information in FAQ

reality-prod:
Updated FAQ to reflect the deprecation of Flash and added information about Firefox for secure web browsing.


diversys:
I think it's just better to drop this section altogether at this point.

